PR #35, opened 2024-09-25
Expose sqlx runtime choice as a feature in the SQL catalog

rdettai:
Not sure whether you are interested in contributions, but I was testing the crate and ran into the following panic when running the README: https://github.com/launchbadge/sqlx/blob/293c55ce896edefc797b6b819c8b27cd327382bb/sqlx-core/src/rt/mod.rs#L143

This PR makes tokio the default runtime, while keeping the possibility to  use `async-std`. But TBH, I doubt many potential users would want to use `async-std` anyway.

### DIFF
--- a/iceberg-sql-catalog/Cargo.toml
+++ b/iceberg-sql-catalog/Cargo.toml
@@ -9,6 +9,10 @@ license = "Apache-2.0"
 
 repository = "https://github.com/JanKaul/iceberg-rust"
 
+[features]
+runtime-tokio = ["sqlx/runtime-tokio"]
+runtime-async-std = ["sqlx/runtime-async-std"]
+
 [dependencies]
 async-trait.workspace = true
 futures.workspace = true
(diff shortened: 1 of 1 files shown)


iceberg-sql-catalog/Cargo.toml
```diff
@@ -9,6 +9,10 @@ license = "Apache-2.0"
 
 repository = "https://github.com/JanKaul/iceberg-rust"
 
+[features]
+runtime-tokio = ["sqlx/runtime-tokio"]
```
JanKaul, line 14: I think it would be good if we make `runtime-tokio` a default feature.

JanKaul:
Thanks a lot for the PR. This is a great point. I think if we make `runtime-tokio` a default feature this is ready to be merged.

rdettai:
Sorry, I had the commit but I forgot to push it.

JanKaul:
Thank you

PR #36, opened 2024-10-01
Fix integration tests

rdettai:
- update s3 config to match latest Trino version
- wait for functional processor node instead of `/v1/info` API


JanKaul:
That's really great, Thank you!

PR #38, opened 2024-10-01
Avoid negative snaphot ids

rdettai:
As the snapshot id is part of the manifest list name, having a negative value gives weird file names:

```
snap--199151010591443303127a4577f-b6fa-49ff-bee4-8d96ad643ac6.avro
```

I assume interpreting the random bytes as u64 was intended to ensure that the value is positive, but the cast to i64 can overflow so some values end up being negative.


JanKaul:
Thanks for catching and fixing that issue.

PR #39, opened 2024-10-01
add prefix to the path of datafiles

### LINKED ISSUE #34: Support object store file layout
I'm migrating a spark workload to datafusion and I'd like to use iceberg tables with the new implementation. One of the things we found helped a lot with the spark / iceberg implementation wrt S3 was the java iceberg's [object store layout](https://iceberg.apache.org/docs/latest/aws/#object-store-file-layout) feature. From what I can see this doesn't seem to be supported here - am I mistaken or is is something that could be included?

JanKaul:
This PR introduces a random prefix into the path of a datafile to avoid throttling of s3 requests.

Closes #34

### DIFF
--- a/iceberg-rust/src/arrow/write.rs
+++ b/iceberg-rust/src/arrow/write.rs
@@ -8,6 +8,7 @@ use futures::{
     SinkExt, StreamExt, TryStreamExt,
 };
 use object_store::{buffered::BufWriter, ObjectStore};
+use std::fmt::Write;
 use std::sync::{
     atomic::{AtomicUsize, Ordering},
     Arc,
@@ -42,7 +43,7 @@ pub async fn write_parquet_partitioned(
     object_store: Arc<dyn ObjectStore>,
     branch: Option<&str>,
 ) -> Result<Vec<DataFile>, ArrowError> {
-    let location = &metadata.location;
+    let data_location = &(metadata.location.clone() + "/data/");
     let schema = metadata.current_schema(branch).map_err(Error::from)?;
     let partition_spec = metadata.default_partition_spec().map_err(Error::from)?;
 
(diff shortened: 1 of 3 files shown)


iceberg-rust/src/arrow/write.rs
```diff
@@ -221,15 +227,21 @@ async fn create_arrow_writer(
...
+        .fold(String::with_capacity(8), |mut acc, x| {
+            write!(&mut acc, "{:02x}", x).unwrap();
+            acc
+        });
 
-    let writer = BufWriter::new(object_store.clone(), parquet_path.clone().into());
+    let parquet_path = strip_prefix(data_location)
+        + &prefix
+        + "/"
+        + partition_path
```
rdettai: 1) I would rather make this optional, most workloads won't benefit from this and you would rather have a nice layout in case you need to inspect your data manually
2) if we use the random prefix, we can get rid of the hive style partition path layout (`/a=10/`), it won't be usable anyway and will just make the key extra long
3) I think `write!(&mut acc, "{:x}", x).unwrap();` should be enough

JanKaul: Thanks for the input, these are great points. I guess making it optional is the best way to go.

I will quickly sketch my reasoning. I would like this library to have good defaults, that users don't have to concern themselves with the partitioning of object storage. And it seems to me that parallelizing the requests with multiple S3 partitions per Iceberg partition has only benefits apart from the ugly file layout. Granted, this might not apply to all use cases but to some it will.

You are right that using a random prefix and the hive style layout probably doesn't make sense. I thought that maybe it would be a tradeoff between the two. But I guess you loose the structure once you introduce the random prefix and there is no point in keeping the partition values.

I will make this configurable with a table property, just like in Java.

PR #40, opened 2024-10-02
Ensure clippy doesn't raise any warnings

rdettai:


Cargo.lock
rdettai, line 1: Not sure why I have a diff on the lock.  Any ideas?

iceberg-rest-catalog/src/lib.rs
```diff
@@ -8,7 +8,9 @@ extern crate serde;
 extern crate serde_json;
 extern crate url;
 
+#[allow(clippy::all)]
 pub mod apis;
 pub mod catalog;
 pub mod error;
+#[allow(clippy::all)]
```
rdettai, line 15: no clippy on generated files

iceberg-rust/src/arrow/partition.rs
```diff
@@ -48,7 +48,8 @@ pub async fn partition_record_batches(
     )>,
     ArrowError,
 > {
-    let (partition_sender, partition_reciever): (
+    #[allow(clippy::type_complexity)]
```
rdettai, line 51: Disabled clippy for now. I guess this was meant to be rewritten anyway to avoid buffering all the batches, no? Do you mind if I add a todo?

iceberg-sql-catalog/src/lib.rs
```diff
@@ -436,6 +436,7 @@ impl Catalog for SqlCatalog {
     async fn update_table(self: Arc<Self>, commit: CommitTable) -> Result<Table, IcebergError> {
         let identifier = commit.identifier;
         let Some(entry) = self.cache.read().unwrap().get(&identifier).cloned() else {
+            #[allow(clippy::if_same_then_else)]
```
rdettai, line 439: I think something is wrong here? we fail if the table metadata is not in cache. Is this cache really necessary? Checking against a relational db is cheap.

iceberg-sql-catalog/src/lib.rs
rdettai, line 1: Noticed that you use string formatting in the SQL queries. This is vulnerable to SQL injections (https://docs.rs/sqlx/latest/sqlx/fn.query.html#dynamic-input-use-query-parameters-prevents-sql-injection). I'll open another PR to use query parameters (`.bind()`) instead.

PR #46, opened 2024-10-24
Improve append

JanKaul:
This PR improves the Append Operation in that it selects only a single manifest file to append the new datafiles to. In case the number of datafiles exceeds a certain threshold, the manifest file is split into smaller manifest files.

This should improve insert performance, as well as the overall structure of the manifest_list/manifest tree.

### DIFF
--- a/datafusion_iceberg/src/table.rs
+++ b/datafusion_iceberg/src/table.rs
@@ -597,10 +597,13 @@ impl DataSink for IcebergDataSink {
 mod tests {
 
     use datafusion::{arrow::array::Int64Array, prelude::SessionContext};
-    use iceberg_rust::spec::{
-        partition::{PartitionField, Transform},
-        schema::Schema,
-        types::{PrimitiveType, StructField, StructType, Type},
+    use iceberg_rust::{
+        catalog::tabular::Tabular,
+        spec::{
+            partition::{PartitionField, Transform},
+            schema::Schema,
+            types::{PrimitiveType, StructField, StructType, Type},
+        },
     };
     use iceberg_rust::{
         catalog::Catalog,
@@ -613,7 +616,7 @@ mod tests {
     };
     use iceberg_sql_catalog::SqlCatalog;
     use object_store::{memory::InMemory, ObjectStore};
-    use std::sync::Arc;
+    use std::{ops::Deref, sync::Arc};
 
     use crate::{catalog::catalog::IcebergCatalog, DataFusionTable};
 
@@ -863,7 +866,7 @@ mod tests {
 
         let ctx = SessionContext::new();
 
-        ctx.register_table("orders", table).unwrap();
+        ctx.register_table("orders", table.clone()).unwrap();
 
         ctx.sql(
             "INSERT INTO orders (id, customer_id, product_id, date, amount) VALUES 
@@ -966,6 +969,13 @@ mod tests {
                 }
             }
         }
+
+        match table.tabular.read().await.deref() {
+            Tabular::Table(table) => {
+                assert_eq!(table.manifests(None, None).await.unwrap().len(), 2);
+            }
+            _ => (),
+        };
     }
 
     #[tokio::test]
--- a/datafusion_iceberg/src/table.rs
+++ b/datafusion_iceberg/src/table.rs
@@ -929,7 +929,17 @@ mod tests {
                 (12, 2, 3, '2020-01-04', 1),
                 (13, 1, 1, '2020-01-05', 4),
                 (14, 3, 2, '2020-01-05', 2),
-                (15, 2, 3, '2020-01-05', 3);",
+                (15, 2, 3, '2020-01-05', 3),
+                (16, 2, 3, '2020-01-05', 3),
+                (17, 1, 3, '2020-01-06', 1),
+                (18, 2, 1, '2020-01-06', 2),
+                (19, 2, 2, '2020-01-06', 1),
+                (20, 1, 2, '2020-01-07', 3),
+                (21, 3, 1, '2020-01-07', 2),
+                (22, 2, 3, '2020-01-07', 1),
+                (23, 1, 1, '2020-01-08', 4),
+                (24, 3, 2, '2020-01-08', 2),
+                (25, 2, 3, '2020-01-08', 3);",
         )
         .await
         .expect("Failed to create query plan for insert")
@@ -961,9 +971,9 @@ mod tests {
                 );
                 for (product_id, amount) in product_ids.iter().zip(amounts) {
                     match product_id.unwrap() {
-                        1 => assert_eq!(amount.unwrap(), 7),
-                        2 => assert_eq!(amount.unwrap(), 4),
-                        3 => assert_eq!(amount.unwrap(), 1),
+                        1 => assert_eq!(amount.unwrap(), 11),
+                        2 => assert_eq!(amount.unwrap(), 7),
+                        3 => assert_eq!(amount.unwrap(), 2),
                         _ => panic!("Unexpected order id"),
                     }
                 }
--- a/datafusion_iceberg/src/table.rs
+++ b/datafusion_iceberg/src/table.rs
@@ -920,7 +920,13 @@ mod tests {
             "INSERT INTO orders (id, customer_id, product_id, date, amount) VALUES 
                 (7, 1, 3, '2020-01-03', 1),
                 (8, 2, 1, '2020-01-03', 2),
-                (9, 2, 2, '2020-01-03', 1);",
+                (9, 2, 2, '2020-01-03', 1),
+                (10, 1, 2, '2020-01-04', 3),
+                (11, 3, 1, '2020-01-04', 2),
+                (12, 2, 3, '2020-01-04', 1),
+                (13, 1, 1, '2020-01-05', 4),
+                (14, 3, 2, '2020-01-05', 2),
+                (15, 2, 3, '2020-01-05', 3);",
         )
         .await
         .expect("Failed to create query plan for insert")
@@ -952,8 +958,9 @@ mod tests {
                 );
                 for (product_id, amount) in product_ids.iter().zip(amounts) {
                     match product_id.unwrap() {
-                        1 => assert_eq!(amount.unwrap(), 3),
-                        2 | 3 => assert_eq!(amount.unwrap(), 1),
+                        1 => assert_eq!(amount.unwrap(), 7),
+                        2 => assert_eq!(amount.unwrap(), 4),
+                        3 => assert_eq!(amount.unwrap(), 1),
                         _ => panic!("Unexpected order id"),
                     }
                 }
--- a/datafusion_iceberg/src/table.rs
+++ b/datafusion_iceberg/src/table.rs
@@ -980,11 +980,8 @@ mod tests {
             }
         }
 
-        match table.tabular.read().await.deref() {
-            Tabular::Table(table) => {
-                assert_eq!(table.manifests(None, None).await.unwrap().len(), 2);
-            }
-            _ => (),
+        if let Tabular::Table(table) = table.tabular.read().await.deref() {
+            assert_eq!(table.manifests(None, None).await.unwrap().len(), 2);
         };
     }
 
--- a/iceberg-rust-spec/src/spec/values.rs
+++ b/iceberg-rust-spec/src/spec/values.rs
@@ -2,13 +2,14 @@
  * Value in iceberg
  */
 
+use core::panic;
 use std::{
     any::Any,
     collections::{btree_map::Keys, BTreeMap, HashMap},
     fmt,
     hash::{DefaultHasher, Hash, Hasher},
     io::Cursor,
-    ops::{Add, Sub},
+    ops::Sub,
     slice::Iter,
 };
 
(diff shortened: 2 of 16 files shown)


iceberg-rust/src/util/mod.rs
rdettai, line 1: This is easy to unit test and I think it is worth it because the implem of `cmp_with_priority` is not straightforward

iceberg-rust/src/table/transaction/append.rs
```diff
@@ -0,0 +1,84 @@
...
+use smallvec::{smallvec, SmallVec};
+
+use crate::{
+    error::Error,
+    util::{cmp_dist, struct_to_smallvec, sub, Rectangle},
+};
+
+/// Split sets of datafiles depending on their partition_values
+#[allow(clippy::type_complexity)]
+pub(crate) fn split_datafiles_once(
```
rdettai: this can be private (helps understanding quickly that this is just a helper for the recursion)

iceberg-rust/Cargo.toml
```diff
@@ -30,4 +30,5 @@ thrift = { version = "0.17.0", default-features = false }
 thiserror = { workspace = true }
 derive-getters = { workspace = true }
 iceberg-rust-spec = { path = "../iceberg-rust-spec", version = "0.5.8" }
+smallvec = { version = "1.13.2", features = ["const_generics"] }
```
rdettai, line 33: smallvec is an interesting optimisation, but:
- do we really need it? Are these allocations on the critical path?
- `SmallVec<[T; 4]` hurts readability (the 4 appears in the function signatures when it's just an optimization detail). If the `smallvec` optim is really useful, type alias it to `SmallVec<T>`

iceberg-rust/src/util/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+use iceberg_rust_spec::{
+    manifest_list::FieldSummary,
+    values::{Struct, TryAdd, TrySub, Value},
+};
+use smallvec::SmallVec;
+
+use crate::error::Error;
+
+#[derive(Clone, PartialEq, Eq)]
+pub(crate) struct Rectangle<C> {
```
rdettai: this is more a cube than a rectangles :smile:

iceberg-rust/src/util/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+
+pub(crate) fn struct_to_smallvec(
+    s: &Struct,
+    names: &SmallVec<[&str; 4]>,
+) -> Result<SmallVec<[Value; 4]>, Error> {
+    names
+        .iter()
+        .map(|x| {
+            s.get(x)
+                .and_then(|x| identity(x).clone())
```
rdettai: what is identity() used for?

iceberg-rust/src/table/transaction/append.rs
```diff
@@ -0,0 +1,84 @@
...
+    names: &SmallVec<[&str; 4]>,
+) -> Result<[(Vec<ManifestEntry>, Rectangle<Value>); 2], Error> {
+    let mut smaller = Vec::new();
+    let mut larger = Vec::new();
+    let mut smaller_rect = None;
+    let mut larger_rect = None;
+
+    for manifest_entry in files {
+        let manifest_entry = manifest_entry?;
+        let position = struct_to_smallvec(manifest_entry.data_file().partition(), names)?;
```
rdettai: this helper is more confusing that actually inlining the extraction:
```suggestion
        let partition_struct = manifest_entry.data_file().partition();
        let position: SmallVec<Value> =  names
        .iter()
        .map(|x| partition_struct.get(x).and_then(|x| x.clone()))
        .collect::<Option<_>>()
        .ok_or(Error::InvalidFormat("Partition struct".to_owned()))?:
```

iceberg-rust/Cargo.toml
JanKaul: The type alias is a great idea!

Generally, I don't know the number of partition columns beforehand and I cannot create the according data-structure at compile time. That's why I might need to allocate memory at runtime.
However, the number of partition columns is typically very small, that's why I would like to reserve the according memory on the stack and only spill to the heap if there are more partition columns. That's what smallvec does. It tries to do everything in an array on the stack and allocates additional memory if needed. This way the common case should be faster, but nothing happens in case there are more columns.

iceberg-rust/src/table/transaction/append.rs
JanKaul: Yes, it's a helper function an can be private.

iceberg-rust/src/util/mod.rs
JanKaul: Ah that's legacy garbage. I tried to just pass the identity function to the and_then and forgot to remove when I needed to clone the value. It should be removed.

iceberg-rust/src/table/transaction/append.rs
JanKaul: Since it's being used multiple times across the crate I would rather put it into a function.

rdettai: ah sory, I missed that it was also used in operation.rs

iceberg-rust/Cargo.toml
rdettai: I agree the optimization makes sense, but I was wondering if it is really necessary? Is the allocation really slowing down the execution? I think this is only the case if you make a huge amount of allocations. We can assume here that it's the case, but I was curious how you got to that conclusion.

When you create abstractions to encapsulate part of your code, you generally want to keep only the minimal information in the signatures to keep them as readable as possible. Here, having a specific `Vec` type adds noise. Aliasing to `Vec4` doesn't really help, because the `4` is an optimization value.

iceberg-rust/src/table/transaction/append.rs
```diff
@@ -0,0 +1,85 @@
...
+            larger,
+            larger_rect.ok_or(Error::NotFound("Upper".to_owned(), "rectangle".to_owned()))?,
+        ),
+    ])
+}
+
+pub(crate) fn split_datafiles(
+    files: impl Iterator<Item = Result<ManifestEntry, Error>>,
+    rect: Rectangle,
+    names: &SmallVec<[&str; 4]>,
```
rdettai: `&[&str]` is enough

iceberg-rust/Cargo.toml
rdettai: I would be tempted to have a type like `PartitionVec` that is used across the entire crate. That would semantically indicate that we are manipulating values relative to partitions, and we could use it to encapsulate the SmallVec<T; 4> optimization. What do you think?

iceberg-rust/src/table/transaction/append.rs
```diff
@@ -0,0 +1,87 @@
...
+    ])
+}
+
+/// Splits the datafiles *n_split* times to decrease the number of datafiles per maniefst. 1 split returns 2 outputs vectors, 2 splits return 4, 3 splits return 8 and so on.
+pub(crate) fn split_datafiles(
+    files: impl Iterator<Item = Result<ManifestEntry, Error>>,
+    rect: Rectangle,
+    names: &[&str],
+    n_split: u32,
+) -> Result<SmallVec<[Vec<ManifestEntry>; 2]>, Error> {
```
rdettai:
```suggestion
) -> Result<Vec<Vec<ManifestEntry>>, Error> {
```

The optimization is not necessary here as `split_datafiles` is called once per append, and using `Vec` is substantially more readable.

iceberg-rust/src/table/transaction/append.rs
```diff
@@ -0,0 +1,87 @@
...
+use smallvec::{smallvec, SmallVec};
+
+use crate::{
+    error::Error,
+    util::{cmp_with_priority, partition_struct_to_vec, try_sub, Rectangle},
+};
+
+/// Split sets of datafiles depending on their partition_values
+#[allow(clippy::type_complexity)]
+pub fn split_datafiles_once(
```
rdettai: you made it even more public :sweat_smile:

iceberg-rust/src/table/transaction/append.rs
```diff
@@ -0,0 +1,87 @@
...
+            smaller_rect.ok_or(Error::NotFound("Lower".to_owned(), "rectangle".to_owned()))?,
+        ),
+        (
+            larger,
+            larger_rect.ok_or(Error::NotFound("Upper".to_owned(), "rectangle".to_owned()))?,
+        ),
+    ])
+}
+
+/// Splits the datafiles *n_split* times to decrease the number of datafiles per maniefst. 1 split returns 2 outputs vectors, 2 splits return 4, 3 splits return 8 and so on.
```
rdettai:
```suggestion
/// Splits the datafiles `n_split` times to decrease the number of datafiles per manifest. Returns `2^n_split` lists of manifest entries.
/// Each split tries to keep identical or close partitions together to improve the pruning performances at the manifest list level.
```

iceberg-rust/src/table/transaction/append.rs
```diff
@@ -0,0 +1,87 @@
...
+pub(crate) fn split_datafiles(
```
rdettai, line 66: Ideally, this should be unit tested as well.

iceberg-rust-spec/src/spec/values.rs
```diff
@@ -817,6 +829,95 @@ mod datetime {
     }
 }
 
+pub trait TryAdd: Sized {
```
rdettai: `try_add` isn't used, I recommend removing it (YAGNI)

iceberg-rust/src/util/mod.rs
```diff
@@ -0,0 +1,421 @@
...
+            if node[i] < self.min[i] {
+                self.min[i] = node[i].clone();
+            }
+            if node[i] > self.max[i] {
+                self.max[i] = node[i].clone();
+            }
+        }
+    }
+
+    /// Determine of one rectangle is larger than the other. Values the earlier columns more than the later.
```
rdettai:
```suggestion
    /// Determine if one rectangle is larger than the other.
    ///
    /// Values in earlier columns weigh more than the later.
```

iceberg-rust-spec/src/spec/values.rs
```diff
@@ -817,6 +829,95 @@ mod datetime {
...
+            (Value::TimestampTZ(own), Value::TimestampTZ(other)) => {
+                Ok(Value::TimestampTZ(own - other))
+            }
+            (Value::String(own), Value::String(other)) => {
+                Ok(Value::LongInt(sub_string(own, other) as i64))
+            }
+            (x, y) => Err(Error::Type(
+                x.datatype().to_string(),
+                y.datatype().to_string(),
+            )),
```
rdettai, line 889: We are still erroring on some valid column types (`UUID`, `Fixed`....). If it's out of scope to handle those in this PR, a TODO would be nice
```suggestion
            // TODO handle missing types (UUID, Fixed...)
            (x, y) => Err(Error::Type(
                x.datatype().to_string(),
                y.datatype().to_string(),
            )),
```

rdettai: That being said, I would still support those types without erroring by implementing a dummy sub of the hash (with a disclaimer that this might be improved)

iceberg-rust/src/table/transaction/operation.rs
rdettai, line 1: `Operation.execute()` is too big, which makes it hard to read the change. Could you refactor it?

datafusion_iceberg/src/table.rs
```diff
@@ -920,7 +920,13 @@ mod tests {
...
                 (7, 1, 3, '2020-01-03', 1),
                 (8, 2, 1, '2020-01-03', 2),
-                (9, 2, 2, '2020-01-03', 1);",
+                (9, 2, 2, '2020-01-03', 1),
+                (10, 1, 2, '2020-01-04', 3),
+                (11, 3, 1, '2020-01-04', 2),
+                (12, 2, 3, '2020-01-04', 1),
+                (13, 1, 1, '2020-01-05', 4),
+                (14, 3, 2, '2020-01-05', 2),
+                (15, 2, 3, '2020-01-05', 3);",
```
rdettai: Are these extra rows meant to trigger the manifest split code path? Either way, I think there should be a test with a large append somewhere that does generate multiple manifest files. It can be here if it's the easiest, but we should also assert that the manifest files were split as expected.

JanKaul: Yes, exactly. It triggers splitting the manifest. I'll insert an assert.

iceberg-rust/src/table/transaction/append.rs
JanKaul: Creating mocked data for a unit test is quite difficult here. I expanded a datafusion test to cover the manifest splitting case.

iceberg-rust/src/util/mod.rs
rdettai: I think what you committed still has some typos: https://github.com/JanKaul/iceberg-rust/blob/4cc2bff2f316d1b08d718afb76483604c23d06e8/iceberg-rust/src/util/mod.rs#L54-L56

PR #48, opened 2024-11-04
Move manifest list reader from spec crate

rdettai:
The manifest list reader should be in the same crate as the manifest file reader (main iceberg-rust).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2473,7 +2473,6 @@ dependencies = [
  "getrandom",
  "itertools 0.10.5",
  "murmur3",
- "object_store",
  "ordered-float 4.2.0",
  "rust_decimal",
  "serde",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ serde_derive = "^1.0"
 serde_json = "^1.0"
 futures = "0.3.30"
 async-trait = "0.1"
-chrono = { version = "0.4", default-features = false, features = ["serde"] }
+chrono = { version = "0.4", default-features = false, features = ["serde", "clock"] }
 arrow = "52.1.0"
 arrow-schema = "52.1.0"
 datafusion = "40.0.0"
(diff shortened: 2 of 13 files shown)


rdettai:
whaou, super weird, removing the dep to object_store somehow breaks the compilation in various places...

JanKaul:
I wonder if moving the manifest file reader out of the `iceberg-rust-spec` crate was premature, as the readers and writers are somewhat linked to the spec. I like the `manifests` method of the snapshot as it makes it obvious where to get the list of manifests from. Maybe we can move the manifest file reader back to the `iceberg-rust-spec` crate but create a separate file for it, so that it's not convoluted with the `manifest_entry`. And similarly with the `ManifestListReader`.

Let me know what you think.

rdettai:
It's nice to have the `iceberg-rust-spec` really focus on the spec: define how the files should look like and help you build valid Iceberg entities. I see a lot of benefit from decoupling it from object `object_store`: re-usability, readability, testability... But on the other hand I also think that having a higher level reader/writer that is `object_store` specific is very useful. Maybe some of the reading/writing logic can maybe be brought back, to have a more powerful spec API, but I would also have an object store specific reader/writer somewhere else.

JanKaul:
And there is the question about the naming scheme. So far I've called the reader of the manifest list "ManifestListReader" and the reader of the manifest "ManifestReader". If you would like to rename "ManifestReader" to "ManifestFileReader", it kind of breaks the naming scheme.
I'm not particularly opposed to it, I just think that there should be a universal naming scheme.

rdettai:
> So far I've called the reader of the manifest list "ManifestListReader" and the reader of the manifest "ManifestReader". If you would like to rename "ManifestReader" to "ManifestFileReader", it kind of breaks the naming scheme.

Right, I forgot to rename some variables here and there (`manifest_reader` -> `manifest_file_reder`). Are you referring to this or some other place(s) where the naming scheme is broken?

rdettai:
After a discussion on the naming convention, we agreed that it's better to stick with Manifest for manifest files and ManifestList for manifest list files.

JanKaul (approved):
Looks good to me. Thanks for taking this on

PR #54, opened 2024-11-09
Add makefile

rdettai:
It is convinient to have a Makefile that shows the developer commands / git hooks that are expected to be used and that are also executed by the CI.


JanKaul (approved):
Thank you!

PR #64, opened 2024-11-18
Refactor readers

rdettai:
Refactor readers to factorize code and simplify operations logic

### DIFF
--- a/iceberg-rust/src/table/manifest_list.rs
+++ b/iceberg-rust/src/table/manifest_list.rs
@@ -2,25 +2,29 @@
  * Helpers to deal with manifest lists and files
 */
 
+use apache_avro::{types::Value as AvroValue, Reader as AvroReader, Schema as AvroSchema};
+use iceberg_rust_spec::{
+    manifest::{partition_value_schema, ManifestEntry},
+    manifest_list::{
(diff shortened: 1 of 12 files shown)


iceberg-rust/src/table/manifest_list.rs
```diff
@@ -57,20 +61,301 @@ impl<'a, 'metadata, R: Read> ManifestListReader<'a, 'metadata, R> {
...
+    /// Create new manifests for each group of provided manifest entries and append them to the manifest list
+    pub async fn append_manifest_entries_to_new_manifests(
+        &mut self,
+        grouped_manifest_entries: Vec<Vec<ManifestEntry>>,
+    ) -> Result<(), Error> {
+        for entry_group in grouped_manifest_entries {
+            let manifest_location = format!(
+                "{}/metadata/{}-m{}.avro",
+                self.table_metadata.location, self.snapshot_uuid, self.manifest_index
+            );
```
JanKaul, line 143: I think the error is coming from here. The manifest_index should be different for every manifest_group. It's overwriting the previous files.

rdettai: Ah yes, looks like I was forgetting to update manifest_index. It should be easy to add an extra test to also catch this.

JanKaul:
Closed in favor of #193

PR #111, opened 2025-02-05
Extend support for parsing decimal types and values

gruuya:
This PR extends support for decimal types, enabling full TPC-H benchmarking with this crate.

The changes include:
- Fix for deserializing a decimal type with a space before the scale (e.g. `decimal(15, 2)`), as is output by pyiceberg
- Parsing of the byte value to the corresponding decimal stat value


JanKaul (approved):
This is great! Thanks for the contribution.

PR #123, opened 2025-02-12
Bump reqwest-middleware

### LINKED ISSUE #122: `iceberg-rest-catalog` offers no way to opt-out of `openssl`
We have a `deny.toml` which prohibits the use of `openssl(-sys)` in favor of `rustls`. However, the crate `iceberg-rest-catalog` uses a version of `reqwest-middleware` crate which includes `openssl` by default, and has no features to switch it off.

There are two possible solutions here:
- Bump `reqwest-middleware` to a version which allows specifying a transitory `rustls-tls` for the `reqwest` crate
- Get rid of `reqwest-middleware` altogether, as it seems it isn't really needed in the catalog implementation

gruuya:
Closes #122.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1999,7 +1999,7 @@ dependencies = [
  "object_store",
  "pin-project-lite",
  "regex",
- "reqwest 0.11.27",
+ "reqwest",
  "serde_json",
  "sqlx",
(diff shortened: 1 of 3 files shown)


JanKaul:
Closed in favor #124

PR #136, opened 2025-02-25
Fix truncate of unpartitioned tables

### LINKED ISSUE #135: Error when truncating unpartitioned table
## Steps to reproduce

```
// Using iceberg-rust at commit 86addf10f5d4c5b93cd735525f40bd547c038183
// table is an unpartitioned table
let table_transaction = TableTransaction::new(table, None).overwrite(vec![]);
table_transaction.commit().await?;
```

## Expected behavior
Should create a new table snapshot with zero data files

## Observed behavior
`commit()` fails with error `Bounding partition values not found`

SergeiPatiakin:
This PR fixes https://github.com/JanKaul/iceberg-rust/issues/135

An error occurs in the calculation of bounding_partition_values if there are no data files.
This PR moves the calculation to a branch that is only entered if we have data files.

This means Operation::Replace can be successfully executed with zero data files, enabling table truncate

### DIFF
--- a/iceberg-rust/src/table/transaction/operation.rs
+++ b/iceberg-rust/src/table/transaction/operation.rs
@@ -395,18 +395,6 @@ impl Operation {
                     .map(|x| x.name())
                     .collect::<SmallVec<[_; 4]>>();
 
-                let bounding_partition_values = files
-                    .iter()
-                    .try_fold(None, |acc, x| {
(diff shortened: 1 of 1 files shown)


JanKaul:
Thanks a lot for the PR and sorry for the late reply.

The PR looks good to me.

Could you rebase again. Then the warnings should be resolved.

SergeiPatiakin:
@JanKaul done

JanKaul:
Thanks again

Issue #117, opened 2025-02-11
Introduce object_store field in Table struct


gruuya:
The REST spec [stipulates](https://github.com/apache/iceberg/blob/20b18acb77847153c743bbd63e799f5c61a71d82/open-api/rest-catalog-open-api.yaml#L923-L925) that 

>  The response contains both configuration and table metadata. The configuration, if non-empty is used
        as additional configuration for the table that overrides catalog configuration. For example, this
        configuration may change the FileIO implementation to be used for the table.

While the response has the corresponding configuration
https://github.com/JanKaul/iceberg-rust/blob/7e707ebae4cbd65c5b0b01e5ab216b1bf1ab95fe/catalogs/iceberg-rest-catalog/src/models/load_table_result.rs#L24
this gets [discarded](https://github.com/JanKaul/iceberg-rust/blob/7e707ebae4cbd65c5b0b01e5ab216b1bf1ab95fe/catalogs/iceberg-rest-catalog/src/catalog.rs#L284) when actually building the table.

Consequently, the tables can only use the "default" object store that comes pre-configured with the catalog, thus reducing flexibility.

gruuya:
This would entail that `Table`/`View`/`MaterializedView` get a `Arc<dyn ObjectStore>` field, constructed either from the provided config options or from the object store builder in the catalog instance.

Besides that, I'm also wondering whether it would make sense then for each of the tabulars to lose the `catalog` field, since one of it's uses is to construct the object store 
https://github.com/JanKaul/iceberg-rust/blob/7e707ebae4cbd65c5b0b01e5ab216b1bf1ab95fe/iceberg-rust/src/table/mod.rs#L85-L89
This would require supplying a `Arc<dyn Catalog>` as an argument in each remaining tabular function that requires one.

JanKaul:
I think it makes sense to introduce an `object_store` field in `Table/View/Matview`. However, I think that a table is always closely related to its catalog and I would keep the catalog field. Another use of the catalog field is to commit the transaction to the catalog using the `update_table` method. 
If the table doesn't hold a reference to its catalog, you would always need to pass a catalog to a transaction to be able to commit it. Logically I don't see any use cases where you would load a table from one catalog and than commit the transaction on another catalog. Not to mention the atomicity of the operations to the metadata.json file.

[...]

PR #144, opened 2025-03-19
Configurable access delegation for RestCatalog

SergeiPatiakin:
Move x_iceberg_access_delegation into RestCatalog `Configuration` struct.
This allows a RestCatalog instance to make requests with the `X-Iceberg-Access-Delegation` header.

### DIFF
--- a/catalogs/iceberg-rest-catalog/src/apis/catalog_api_api.rs
+++ b/catalogs/iceberg-rest-catalog/src/apis/catalog_api_api.rs
@@ -782,11 +782,10 @@ pub async fn load_table(
     prefix: Option<&str>,
     namespace: &str,
     table: &str,
-    x_iceberg_access_delegation: Option<&str>,
     snapshots: Option<&str>,
 ) -> Result<models::LoadTableResult, Error<LoadTableError>> {
     let mut headers = HashMap::new();
-    if let Some(param_value) = x_iceberg_access_delegation {
+    if let Some(param_value) = &configuration.x_iceberg_access_delegation {
         headers.insert(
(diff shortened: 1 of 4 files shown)


JanKaul:
Looks like a good idea. Thanks for the PR.

JanKaul:
Shall I go ahead and merge this or do you want to extend the draft?

SergeiPatiakin:
I'm a bit stuck with this one. My idea was to let consumers of `struct RestCatalog` use the vended credentials returned from the REST API in `LoadTableResult` or `CreateTableResult` for object store access. However I don't see a way to make that happen with the current interface of `struct RestCatalog` and `trait Catalog`. @JanKaul do you have any suggestion?

Maybe add a method like this? Not very clean though:
```
async fn RestCatalog::load_tabular_and_credentials(self: Arc<Self>, identifier: &Identifier) -> Result<(Tabular, Vec<StorageCredential>), Error>
```

JanKaul:
Do you need the actual `Vec<StorageCredential>`? Or would it suffice to pass an `Arc<dyn ObjectStore>` configured with the provided `StorageCredentials`?

I think https://github.com/JanKaul/iceberg-rust/issues/117 goes into a similar direction where the `ObjectStore` is passed as a field of `Table`/`Tabular`. Would this solve your problem? I think this idea would have a lot of merit.

SergeiPatiakin:
> Or would it suffice to pass an Arc<dyn ObjectStore> configured with the provided StorageCredentials?

Yes, that would suffice for me. Upvote for #117

PR #145, opened 2025-03-20
chore: Update DataFusion to v46.0.0

lizardoluis:
This PR upgrades DataFusion to its latest version.


datafusion_iceberg/src/table.rs
```diff
@@ -478,7 +478,15 @@ async fn table_scan(
...
     });
 
-    // Create plan for every parition with delete files
+    let file_source = if let Some(physical_predicate) = physical_predicate.clone() {
+        Arc::new(
+            ParquetSource::default().with_predicate(Arc::clone(&file_schema), physical_predicate),
+        )
+    } else {
+        Arc::new(ParquetSource::default())
+    };
```
lizardoluis: Please double check if this is the proper way of creating this file_source.

datafusion_iceberg/src/table.rs
```diff
@@ -561,17 +572,25 @@ async fn table_scan(
...
-                            };
+                            let delete_file_source =
+                                if let Some(physical_predicate) = physical_predicate.clone() {
+                                    Arc::new(ParquetSource::default().with_predicate(
+                                        Arc::clone(&delete_file_schema),
+                                        physical_predicate,
+                                    ))
+                                } else {
+                                    Arc::new(ParquetSource::default())
+                                };
```
lizardoluis: Check here too please

datafusion_iceberg/src/table.rs
JanKaul: From what I understand the ParquetSource looks good.

Now that I'm looking into it, what do you think about setting [with_pushdown_filters](https://docs.rs/datafusion/latest/datafusion/datasource/physical_plan/parquet/source/struct.ParquetSource.html#method.with_pushdown_filters) to true. It's false by default.

It should improve the performance.

datafusion_iceberg/src/table.rs
JanKaul: Similar to the above, I would activate the filter pushdown.

JanKaul:
Thanks a lot for the PR.

datafusion_iceberg/src/table.rs
lizardoluis: Added to it

datafusion_iceberg/src/table.rs
lizardoluis: Added to it

PR #147, opened 2025-03-21
Fix negative return value for RestCatalog::tabular_exists

SergeiPatiakin:
With this PR, `RestCatalog::tabular_exists` returns `Ok(false)` if the tabular doesn't exist and no error occurred


SergeiPatiakin:
@JanKaul I made an attempt at writing test cases but faced flakiness (#148 ) in test setup and couldn't run my test cases. Please check if they pass for you.

JanKaul:
Looks good to me. Thanks for the PR.

PR #149, opened 2025-03-24
Refactor/object store in table

### LINKED ISSUE #117: Introduce object_store field in Table struct

JanKaul:
Closes #117 

This PR creates an `object_store` field for Tabular structs to have a reference to the underlying ObjectStore. Before this was part of the catalog trait. This simplifies setting up the object_store with credentials for individual tables.


gruuya (commented):
Thank you, overall I think this is improves the flexibility of `Catalog`/`Tabular` objects!

That said, I still suspect that `Catalog`s should not have the `object_store_builder` field, at least not the `RestCatalog` (or it should be made into a trait). This is because, as clients of the catalog we really don't need to know the connection options for the `object_store` ahead of time (or at all)—all the adequate options should be returned within the load/create tabular responses. 

In other words, we only need to know the namespace/table identifier (and warehouse/token), and the tables `object_store` options will come from the catalog server. 

With the present changes we still need to pre-populate the `ObjectStoreBuilder` with the correct options, and only the bucket gets set upon finding the table location (by calling `ObjectStoreBuilder::build`).

JanKaul:
You're right, the REST catalog shouldn't have an `object_store_builder`.

This is supposed to be a refactor PR to lay the groundwork for any further changes. I'll update the REST catalog with credential vending in a separate PR.

PR #150, opened 2025-03-24
Fix/rest catalog test

JanKaul:
This PR fixes the issues with the catalog tests where the localstack bucket hasn't been created in time for the test to run.


SergeiPatiakin:
I think the test failure is due to a flaw in PR https://github.com/JanKaul/iceberg-rust/pull/147 . Feel free to revert that PR, I will re-implement now that we have a fix for the test issue.

JanKaul:
The idea of PR #147 is correct. Just the passing of the correct error types was lacking beforehand. I fixed that in the last commit here.

PR #154, opened 2025-03-31
Avoid removing snapshots and data files for Operation::Replace

SergeiPatiakin:
Immediately removing snapshots and data files may disrupt concurrent readers who are still reading from the previous snapshot


JanKaul:
Thanks for bringing this up. You're right, the current implementation is too eager in deleting files. After reading into this topic, Iceberg only deletes the old files when an "ExpireSnaphost" operation is performed on the old snapshot. So this change definitely corrects the behavior of the "Replace" Operation.

Thanks for the PR.

PR #159, opened 2025-04-15
Add support for `s3a` and `gs` prefixes in `Bucket::from_path`

lizardoluis:
This PR enhances the `Bucket::from_path` function to support additional prefixes for identifying cloud storage buckets. Specifically, it adds support for:  
- `s3a://` as an alternative to `s3://` for AWS S3 buckets.  
- `gs://` as an alternative to `gcs://` for Google Cloud Storage buckets.


JanKaul:
That's great. Thanks for your PR!

PR #160, opened 2025-04-22
fix: set the correct snapshot number for the replace operation output

gruuya:
Instead of the hard-coded `0` use the calculated sequence number (`sequence_number = table_metadata.last_sequence_number + 1`) for the upcoming snapshot's sequence number in the replace operation output.

Besides the spec saying this should be a monotonically increasing number, there are also a number of [validations](https://github.com/apache/iceberg-rust/pull/1167) for this which means this operation would error out with something like `Invalid snapshot with id 8726289446966218055 and sequence number 1 greater than last sequence number 0` otherwise.


JanKaul:
Thanks for the PR. At the time I thought that since you logically replace the whole data, your sequence number starts at 0 again. But this looks better.

PR #163, opened 2025-04-26
fix: remove `distinct_counts` manifest field

### LINKED ISSUE #162: `distinct_counts` manifest field is deprecated and conflicts with pyiceberg reads
Whilst this field has been an optional field in v1/v2 it has been officially deprecated in v3: https://lists.apache.org/thread/j52tsojv0x4bopxyzsp7m7bqt23n5fnd.

However it seems it has been unofficially deprecated for much longer
- https://github.com/apache/iceberg/issues/767
- https://github.com/apache/iceberg/pull/12182
- https://github.com/apache/iceberg-rust/pull/79#discussion_r1425045199

The last one is especially pertinent since pyiceberg doesn't expect the field in the avro manifest schema https://github.com/apache/iceberg-python/blob/be528ae2a936e9c4432b48ff3370a682cd9029f1/pyiceberg/manifest.py#L114-L292

Consequently when trying to read a table written by this crate that has the `distinct_counts` field in the schema the following cryptic error gets thrown 
```python
Traceback (most recent call last):
  File "/Users/gruuya/EDB/test_read_catalog.py", line 105, in <module>
    print(table.scan(limit=1).to_arrow())
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/gruuya/dev/iceberg-python/pyiceberg/table/__init__.py", line 1792, in to_arrow
    ).to_table(self.plan_files())
               ^^^^^^^^^^^^^^^^^
  File "/Users/gruuya/dev/iceberg-python/pyiceberg/table/__init__.py", line 1741, in plan_files
    for manifest_entry in chain(
                          ^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/concurrent/futures/_base.py", line 619, in result_iterator
    yield _result_or_cancel(fs.pop())
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/concurrent/futures/_base.py", line 317, in _result_or_cancel
    return fut.result(timeout)
           ^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/concurrent/futures/_base.py", line 456, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/gruuya/dev/iceberg-python/pyiceberg/table/__init__.py", line 1743, in <lambda>
    lambda args: _open_manifest(*args),
                 ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/gruuya/dev/iceberg-python/pyiceberg/table/__init__.py", line 1592, in _open_manifest
    for manifest_entry in manifest.fetch_manifest_entry(io, discard_deleted=True)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/gruuya/dev/iceberg-python/pyiceberg/manifest.py", line 708, in fetch_manifest_entry
    with AvroFile[ManifestEntry](
  File "/Users/gruuya/dev/iceberg-python/pyiceberg/avro/file.py", line 184, in __enter__
    self.reader = resolve_reader(self.schema, self.read_schema, self.read_types, self.read_enums)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/gruuya/dev/iceberg-python/pyiceberg/avro/resolver.py", line 253, in resolve_reader
    return visit_with_partner(file_schema, read_schema, ReadSchemaResolver(read_types, read_enums), SchemaPartnerAccessor())  # type: ignore
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/functools.py", line 909, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/gruuya/dev/iceberg-python/pyiceberg/schema.py", line 662, in _
    return visitor.schema(schema, partner, visit_with_partner(schema.as_struct(), struct_partner, visitor, accessor))  # type: ignore
                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/functools.py", line 909, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/gruuya/dev/iceberg-python/pyiceberg/schema.py", line 672, in _
    field_result = visit_with_partner(field.field_type, field_partner, visitor, accessor)  # type: ignore
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/functools.py", line 909, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/gruuya/dev/iceberg-python/pyiceberg/schema.py", line 672, in _
    field_result = visit_with_partner(field.field_type, field_partner, visitor, accessor)  # type: ignore
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/functools.py", line 909, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/gruuya/dev/iceberg-python/pyiceberg/schema.py", line 694, in _
    key_partner = accessor.map_key_partner(partner)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/gruuya/dev/iceberg-python/pyiceberg/avro/resolver.py", line 536, in map_key_partner
    raise ResolveError(f"File/read schema are not aligned for map, got {partner_map}")
pyiceberg.exceptions.ResolveError: File/read schema are not aligned for map, got None
```

One potential resolution is completely removing this field in both `DataFileV1`/`DataFileV2`.

gruuya:
Closes #162.

In this PR `distinct_counts` is removed from both `DataFileV1`/`DataFileV2` schema.

I've tested this change manually and pyiceberg reads now work.

### DIFF
--- a/datafusion_iceberg/src/statistics.rs
+++ b/datafusion_iceberg/src/statistics.rs
@@ -126,12 +126,7 @@ fn column_statistics<'a>(
                     ))
                 })
                 .unwrap_or(Precision::Absent),
-            distinct_count: data_file
-                .distinct_counts()
-                .as_ref()
-                .and_then(|x| x.get(&id))
-                .map(|x| Precision::Exact(*x as usize))
-                .unwrap_or(Precision::Absent),
+            distinct_count: Precision::Absent,
             sum_value: Precision::Absent,
         }
     })
--- a/iceberg-rust-spec/src/spec/manifest.rs
+++ b/iceberg-rust-spec/src/spec/manifest.rs
@@ -539,8 +539,6 @@ pub struct DataFile {
     null_value_counts: Option<AvroMap<i64>>,
     /// Map from column id to number of NaN values
     nan_value_counts: Option<AvroMap<i64>>,
-    /// Map from column id to number of distinct values in the column.
-    distinct_counts: Option<AvroMap<i64>>,
     /// Map from column id to lower bound in the column
     lower_bounds: Option<HashMap<i32, Value>>,
     /// Map from column id to upper bound in the column
@@ -584,7 +582,6 @@ impl DataFile {
             value_counts: value.value_counts,
             null_value_counts: value.null_value_counts,
             nan_value_counts: value.nan_value_counts,
-            distinct_counts: value.distinct_counts,
             lower_bounds: value
                 .lower_bounds
                 .map(|map| map.into_value_map(schema.fields()))
@@ -618,7 +615,6 @@ impl DataFile {
             value_counts: value.value_counts,
             null_value_counts: value.null_value_counts,
             nan_value_counts: value.nan_value_counts,
-            distinct_counts: value.distinct_counts,
             lower_bounds: value
                 .lower_bounds
                 .map(|map| map.into_value_map(schema.fields()))
@@ -658,8 +654,6 @@ pub struct DataFileV2 {
     pub null_value_counts: Option<AvroMap<i64>>,
     /// Map from column id to number of NaN values
     pub nan_value_counts: Option<AvroMap<i64>>,
-    /// Map from column id to number of distinct values in the column.
-    pub distinct_counts: Option<AvroMap<i64>>,
     /// Map from column id to lower bound in the column
     pub lower_bounds: Option<AvroMap<ByteBuf>>,
     /// Map from column id to upper bound in the column
@@ -701,8 +695,6 @@ pub struct DataFileV1 {
     pub null_value_counts: Option<AvroMap<i64>>,
     /// Map from column id to number of NaN values
     pub nan_value_counts: Option<AvroMap<i64>>,
-    /// Map from column id to number of distinct values in the column.
-    pub distinct_counts: Option<AvroMap<i64>>,
     /// Map from column id to lower bound in the column
     pub lower_bounds: Option<AvroMap<ByteBuf>>,
     /// Map from column id to upper bound in the column
@@ -728,7 +720,6 @@ impl From<DataFile> for DataFileV2 {
             value_counts: value.value_counts,
             null_value_counts: value.null_value_counts,
             nan_value_counts: value.nan_value_counts,
-            distinct_counts: value.distinct_counts,
             lower_bounds: value.lower_bounds.map(Into::into),
             upper_bounds: value.upper_bounds.map(Into::into),
             key_metadata: value.key_metadata,
@@ -751,7 +742,6 @@ impl From<DataFile> for DataFileV1 {
             value_counts: value.value_counts,
             null_value_counts: value.null_value_counts,
             nan_value_counts: value.nan_value_counts,
-            distinct_counts: value.distinct_counts,
             lower_bounds: value.lower_bounds.map(Into::into),
             upper_bounds: value.upper_bounds.map(Into::into),
             key_metadata: value.key_metadata,
@@ -777,7 +767,6 @@ impl From<DataFileV1> for DataFileV2 {
             value_counts: v1.value_counts,
             null_value_counts: v1.null_value_counts,
             nan_value_counts: v1.nan_value_counts,
-            distinct_counts: v1.distinct_counts,
             lower_bounds: v1.lower_bounds,
             upper_bounds: v1.upper_bounds,
             key_metadata: v1.key_metadata,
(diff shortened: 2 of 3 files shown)


JanKaul:
Hi, thanks for the PR. I was wondering if we could also resolve the issue by setting the values to None/Precision::Absent but leave the field as part of the schema. I somehow would like to stay close to the spec even when we're not using the field. It should then finally be resolved by adding the V3 spec implementation.

It actually worries me that pyiceberg doesn't seem to distinguish between different table versions.

gruuya:
> I was wondering if we could also resolve the issue by setting the values to None/Precision::Absent but leave the field as part of the schema.

I just tried setting this https://github.com/JanKaul/iceberg-rust/blob/4025cbc7974af73f3bb47eb668f341996f866686/iceberg-rust/src/file_format/parquet.rs#L253
to `None`, but it didn't work (pyiceberg read fails with same error).

> I somehow would like to stay close to the spec even when we're not using the field. It should then finally be resolved by adding the V3 spec implementation.

Yeah that makes sense to me.

> It actually worries me that pyiceberg doesn't seem to distinguish between different table versions.

Agreed as well; the thing is this problem can probably be [circumvented](https://github.com/apache/iceberg-python/pull/1613) by constructing the Avro reader manually, but this isn't an option when using the catalog, e.g.
```python
table = catalog.load_table((namespace, table_name))
print(table.scan(limit=1).to_arrow())
```
which presents a big problem to us.

gruuya:
> >     I somehow would like to stay close to the spec even when we're not using the field. It should then finally be resolved by adding the V3 spec implementation.
> 
> Yeah that makes sense to me.

That said, the [java repo doesn't seem to have this field defined](https://github.com/apache/iceberg/blob/dc26b72ad016840b79d62bf8a84b7f2109e9b71b/api/src/main/java/org/apache/iceberg/DataFile.java#L37-L117) either, as if it didn't exist in v1/v2 at all.

So the problem seems to be that the official spec is misaligned with two main implementations (java and python ones), since they effectively stopped using the field long before it got officially deprecated.

JanKaul:
Well, in that case there isn't really a choice here. We should be compatible with the other implementations. Thanks for trying though.

gruuya:
> Well, in that case there isn't really a choice here. We should be compatible with the other implementations. Thanks for trying though.

Thanks!

The root of the problem as stated above is there is a mismatch between the spec and the reference (java) as well as the main supporting (python, apache rust) implementations.

From what I can [tell](https://github.com/JanKaul/iceberg-rust/issues/162#issuecomment-2834417780), the community decided to treat the field as if it never was part of the spec (even though it still formally is in v1/v2), and thus there is a conundrum now.

PR #166, opened 2025-04-29
Avoid unneeded schema change for view updates

SergeiPatiakin:
- If the new schema is the same as the old schema, there is no need to update it
- Where a schema update it needed, order it first before ViewUpdate::AddViewVersion


JanKaul:
Thanks for the PR. Looks good to me!

PR #168, opened 2025-05-05
Upgrade sqlparser to v0.56.0 by pinning v0. Upgrade avro to v0.18

lizardoluis:
`Datafusion` v47 depends of `sqlparser` v0.55 and v0.56. So with this change I removed the dependency to the old v0.54.


JanKaul:
I don't see sqlparser v0.56 as datafusion v47 dependency. But I think it would be great to pin sqlparser v0.55. What do you think?

lizardoluis:
Sounds good. Pushed the change

JanKaul:
Great! Thanks a lot.

PR #176, opened 2025-05-14
feat: Allow reading compressed metadata files

lizardoluis:
The metadata files can be compressed. In this case, they have the suffix `gz.metadata.json`. This PR adds support for reading them by first decompressing.

### DIFF
--- a/iceberg-rust/src/object_store/store.rs
+++ b/iceberg-rust/src/object_store/store.rs
@@ -272,7 +272,6 @@ mod tests {
         assert!(result.is_ok());
         let metadata = result.unwrap();
         if let TabularMetadata::Table(table_metadata) = metadata {
-            // Add specific checks for `table_metadata` fields if needed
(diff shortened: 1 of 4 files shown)


iceberg-rust/src/object_store/store.rs
```diff
@@ -80,9 +83,23 @@ fn version_hint_path(original: &str) -> Option<String> {
...
+        let mut decoder = GzDecoder::new(bytes);
+        let mut decompressed_data = Vec::new();
+        decoder
+            .read_to_end(&mut decompressed_data)
+            .map_err(|e| Error::Decompress(e.to_string()))?;
+        serde_json::from_slice(&decompressed_data).map_err(Error::from)
+    } else {
+        serde_json::from_slice(bytes).map_err(Error::from)
+    }
+}
```
lizardoluis, line 97: I separated this code into this function to allow easy testing.

JanKaul:
Thanks for taking this on. Looks good to me.

PR #178, opened 2025-05-21
feat: upstream object store deduplication

### LINKED ISSUE #161: Datafusion mixes vended credentials between different tables in the same bucket
## Steps To Reproduce
- Create two tables with identical data in an Iceberg REST catalog with x_iceberg_access_delegation = Some("vended-credentials"):
	- `t1` at `s3://some-bucket/t1`
	- `t2` at `s3://some-bucket/t2`
- Run the Datafusion query `SELECT * FROM t1 UNION ALL SELECT * FROM t2`

## Expected behavior
- Query should succeed
- Credentials vended for `t1` should be used for scanning `t1`
- Credentials vended for `t2` should be used for scanning `t2`

## Observed behavior
- Query fails with permission errors
- Credentials vended for `t2` are used for scanning `t1`

## Comments
- The root cause is Datafusion's DefaultObjectStoreRegistry only takes bucket name into account when looking up the key ([register_store](https://github.com/apache/datafusion/blob/main/datafusion/execution/src/object_store.rs#L230), [get_url_key](https://github.com/apache/datafusion/blob/main/datafusion/execution/src/object_store.rs#L248)). This means that the object store instance for `t1` is discarded when the object store instance for `t2` is registered
- We have been able to work around this with https://github.com/splitgraph/iceberg-rs/pull/9 but we're not sure how well this solution plays with other folks' needs. Let me know if you have a better idea for a fix!

gruuya:
Upstream of https://github.com/splitgraph/iceberg-rs/pull/9

Closes #161

### DIFF
--- a/datafusion_iceberg/src/table.rs
+++ b/datafusion_iceberg/src/table.rs
@@ -261,6 +261,25 @@ impl TableProvider for DataFusionTable {
     }
 }
 
+// Create a fake object store URL. Different table paths should produce fake URLs
+// that differ in the host name, because DF's DefaultObjectStoreRegistry only takes
+// hostname into account
+fn fake_object_store_url(table_location_url: &str) -> Option<ObjectStoreUrl> {
+    let mut u = url::Url::parse(&table_location_url).ok()?;
+    u.set_host(Some(&format!(
+        "{}{}",
+        u.host_str().unwrap_or("-"),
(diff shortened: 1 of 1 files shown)


JanKaul:
Thanks a lot for working on this! Looks good to me.

PR #180, opened 2025-05-21
Respect other dialects when updating representation

SergeiPatiakin:


SergeiPatiakin:
Accident, please disregard

PR #181, opened 2025-05-22
Allow updating multiple view representations

SergeiPatiakin:
- Changed `Operation::UpdateRepresentation` to `Operation::UpdateRepresentations` which can take multiple representations
- `Operation::UpdateRepresentations` preserves representations in other dialects instead of overwriting them

This is a breaking change but consumers can easily migrate by changing `update_representation(r)` to `update_representations(vec![r])`


JanKaul:
Thanks for the PR. It's great to generalize this to multiple representations. I didn't think of it at the time.

PR #182, opened 2025-05-27
fix: transparent get_table_metadata_location error

gruuya:
Currently any error whatsoever that occurs during fetching metadata will be displayed as `Entity not found in catalog`, even though it may originate from e.g. connection issues or more likely permission issues.

This PR makes the underlying error transparent by adding another s3tables `GetTableMetadataLocation` error variant, in line with the rest of the client calls.


JanKaul:
I would close this in favor of #184 because the CatalogNotFound error is sometimes being handled at a higher level. Thanks for the PR though!

gruuya:
> I would close this in favor of https://github.com/JanKaul/iceberg-rust/pull/184 because the CatalogNotFound error is sometimes being handled at a higher level.

Agreed, that's a better approach, thanks!

PR #183, opened 2025-05-27
feat: don't set default_catalog in view builders if catalog name is empty

gruuya:


JanKaul:
Great catch, thanks a lot for the PR! Looks good to me.

PR #187, opened 2025-06-04
feat: vend credentials when loading a tabular

gruuya:
When loading a table/view make sure to pass the `vended-credentials` header, so that the catalog server returns proper options for accessing the object store. Note that this is also the default in pyiceberg https://github.com/apache/iceberg-python/blob/a67c5592f3243d255519581fedfcc5d93274b9c8/pyiceberg/catalog/rest/__init__.py#L463

In addition parse a couple more options, such as `s3.endpoint` when creating the object store.

### DIFF
--- a/catalogs/iceberg-rest-catalog/src/catalog.rs
+++ b/catalogs/iceberg-rest-catalog/src/catalog.rs
@@ -1,3 +1,11 @@
+use crate::{
+    apis::{
+        self,
+        catalog_api_api::{self, NamespaceExistsError},
+        configuration::Configuration,
+    },
+    models::{self, StorageCredential},
+};
 use async_trait::async_trait;
 use futures::{FutureExt, TryFutureExt};
 /**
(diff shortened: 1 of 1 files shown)


catalogs/iceberg-rest-catalog/src/catalog.rs
```diff
@@ -613,43 +613,59 @@ const CLIENT_REGION: &str = "client.region";
...
-    };
-
-    let Some(config) = config else {
-        return Ok(None);
+        (Some(credentials), Some(config)) => {
+            // Enrich credentials with other options that might only be found in the config (e.g.
+            // a custom endpoint)
+            let mut options = (&credentials[0].config).clone();
+            options.extend(config.clone());
+            options
```
gruuya: In the particular case of Lakekeeper, `response.storage_credentials[0].config` contains only `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` and `AWS_SESSION_TOKEN`.

OTOH, `response.config` contains those as well, but then in addition it also has things like `CLIENT_REGION`, `AWS_REGION`, `AWS_ENDPOINT`, etc.

gruuya:
Closing in favor of #188, which also adds parsing for GCS options and some tests too.

PR #188, opened 2025-06-05
feat: vend credentials when loading a tabular

gruuya:
When loading a table/view make sure to pass the vended-credentials header, so that the catalog server returns proper options for accessing the object store. Note that this is also the default in pyiceberg https://github.com/apache/iceberg-python/blob/a67c5592f3243d255519581fedfcc5d93274b9c8/pyiceberg/catalog/rest/__init__.py#L463

In addition parse a couple more options, such as s3.endpoint when creating the object store, as well as some GCP equivalent options.

### DIFF
--- a/catalogs/iceberg-rest-catalog/src/catalog.rs
+++ b/catalogs/iceberg-rest-catalog/src/catalog.rs
@@ -8,6 +8,7 @@ use crate::{
 };
 use async_trait::async_trait;
 use futures::{FutureExt, TryFutureExt};
+use iceberg_rust::object_store::parse::object_store_from_config;
 /**
 Iceberg rest catalog implementation
 */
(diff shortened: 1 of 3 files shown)


catalogs/iceberg-rest-catalog/src/catalog.rs
```diff
@@ -609,48 +610,24 @@ impl CatalogList for RestNoPrefixCatalogList {
...
-            builder = builder.with_secret_access_key(secret_access_key)
-        }
-        if let Some(session_token) = session_token {
-            builder = builder.with_token(session_token)
+        (Some(credentials), Some(config)) => {
+            // Enrich credentials with other options that might only be found in the config (e.g.
+            // a custom endpoint)
+            let mut options = credentials[0].config.clone();
+            options.extend(config.clone());
+            options
```
gruuya, line 622: In the particular case of Lakekeeper, `response.storage_credentials[0].config` contains only `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY` and `AWS_SESSION_TOKEN`.

OTOH, response.config contains those as well, but then in addition it also has things like `CLIENT_REGION`, `AWS_REGION`, `AWS_ENDPOINT`, etc.

gruuya:
Hi @JanKaul, can you do a review and see if this makes sense to you?

JanKaul:
Thanks for working on this. LGTM.

PR #201, opened 2025-07-02
Add concurrency for manifest read

SergeiPatiakin:
We have been seeing slow performance compared to pyiceberg for planning with a large number of manifest files with relatively small data files. We found the main difference was reading manifest files from the object store, which is parallelized in pyiceberg: https://github.com/apache/iceberg-python/blob/main/pyiceberg/table/__init__.py#L1817-L1831

This PR adds parallelism to improve planning performance.


SergeiPatiakin:
Closing because seems to be a duplicate of https://github.com/JanKaul/iceberg-rust/pull/190

JanKaul:
Thanks for looking into this. I'm not sure if there is a large difference between out implementations. Please let me know if you still have any issues with the performance. We can also go with your solution in the future.

PR #205, opened 2025-07-09
Polish equality delete test

gruuya:
Make asserts more explicit by using `assert_batches_eq!`.

Also reload the table prior to applying equality deletes, otherwise the deletes are applied to an empty table (the initial snapshot prior to INSERT-ing the data), which was obfuscated by the existing assertion.

### DIFF
--- a/datafusion_iceberg/tests/equality_delete.rs
+++ b/datafusion_iceberg/tests/equality_delete.rs
@@ -1,11 +1,8 @@
-use std::sync::Arc;
-
-use datafusion::{
-    arrow::{array::Int64Array, error::ArrowError},
-    prelude::SessionContext,
-};
+use datafusion::{arrow::error::ArrowError, assert_batches_eq, prelude::SessionContext};
 use datafusion_iceberg::catalog::catalog::IcebergCatalog;
 use futures::stream;
+use iceberg_rust::catalog::identifier::Identifier;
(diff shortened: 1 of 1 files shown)


datafusion_iceberg/tests/equality_delete.rs
```diff
@@ -166,38 +168,20 @@ pub async fn test_equality_delete() {
...
-        .sql("select product_id, sum(amount) from warehouse.test.orders group by product_id;")
+        .sql("select product_id, sum(amount) from warehouse.test.orders group by product_id order by product_id")
         .await
         .expect("Failed to create plan for select")
         .collect()
         .await
         .expect("Failed to execute select query");
 
-    for batch in batches {
-        if batch.num_rows() != 0 {
```
gruuya, line 177: Without the table load above, the equality deletes are applied on top of the initial snapshot (without any rows), and so the end result is that there are no rows in the new snapshot either. Consequently the below assertions were never triggered.

JanKaul:
This is great! Makes the tests much more reliable. Thanks a lot for the PR!

datafusion_iceberg/tests/equality_delete.rs
JanKaul: Great catch! The assert_batch makes this much better than before.

PR #218, opened 2025-07-30
fix: skip adding another plan when no data file groups left during table scan

### LINKED ISSUE #217: `table_scan` execution panic for tables with matching data and delete files
If I change the following https://github.com/JanKaul/iceberg-rust/blob/ceb696f509e982346ea11255a055e6355eb5f500/datafusion_iceberg/tests/equality_delete.rs#L177

to be a `select *` instead of an aggregation, i see the following panic
```
thread 'test_equality_delete' panicked at datafusion_iceberg/tests/equality_delete.rs:182:10:
Failed to execute select query: Context("SanityCheckPlan", Plan("Plan: [\"SortExec: expr=[id@0 ASC NULLS LAST], preserve_partitioning=[false]\", \"  DataSourceExec: file_groups={0 groups: []}, projection=[id, customer_id, product_id, date, amount], file_type=parquet\"] does not satisfy distribution requirements: SinglePartition. Child-0 output partitioning: UnknownPartitioning(0)"))
```

The plan for that looks like this
```
    "| initial_physical_plan                                      | UnionExec                                                                                                                                                                                                                                                                                                      |",
    "|                                                            |   ProjectionExec: expr=[id@0 as id, customer_id@1 as customer_id, product_id@2 as product_id, date@3 as date, amount@4 as amount]                                                                                                                                                                              |",
    "|                                                            |     HashJoinExec: mode=CollectLeft, join_type=RightAnti, on=[(id@0, id@0), (customer_id@1, customer_id@1), (product_id@2, product_id@2), (date@3, date@3)]                                                                                                                                                     |",
    "|                                                            |       DataSourceExec: file_groups={1 group: [[test/orders/data/date_day=18262/64b47434-6d07-11f0-88f8-de51894a27a1.parquet]]}, projection=[id, customer_id, product_id, date, date_day], file_type=parquet                                                                                                     |",
    "|                                                            |       DataSourceExec: file_groups={1 group: [[test/orders/data/date_day=18262/64b02528-6d07-11f0-88f7-6d5546fe4045.parquet]]}, projection=[id, customer_id, product_id, date, amount], file_type=parquet                                                                                                       |",
    "|                                                            |   ProjectionExec: expr=[id@0 as id, customer_id@1 as customer_id, product_id@2 as product_id, date@3 as date, amount@4 as amount]                                                                                                                                                                              |",
    "|                                                            |     HashJoinExec: mode=CollectLeft, join_type=RightAnti, on=[(id@0, id@0), (customer_id@1, customer_id@1), (product_id@2, product_id@2), (date@3, date@3)]                                                                                                                                                     |",
    "|                                                            |       DataSourceExec: file_groups={1 group: [[test/orders/data/date_day=18294/64b48f32-6d07-11f0-88f9-bf3bba452bfc.parquet]]}, projection=[id, customer_id, product_id, date, date_day], file_type=parquet                                                                                                     |",
    "|                                                            |       DataSourceExec: file_groups={1 group: [[test/orders/data/date_day=18294/64af9cb6-6d07-11f0-88f6-1d36c1f3beb3.parquet]]}, projection=[id, customer_id, product_id, date, amount], file_type=parquet                                                                                                       |",
    "|                                                            |   DataSourceExec: file_groups={0 groups: []}, projection=[id, customer_id, product_id, date, amount], file_type=parquet                                                                                                                                                                                        |",
    "|                                                            |                                                                                                                                                                                                                                                                                                                |",
```

That last `DataSourceExec: file_groups={0 groups: []}` is the cause of the problem, and it happens due to the fact that in this test all data file groups have matching equality delete groups, meaning that once all of those are paired up for a (anti) join, there are no more file groups left when constructing the other plan here https://github.com/JanKaul/iceberg-rust/blob/ceb696f509e982346ea11255a055e6355eb5f500/datafusion_iceberg/src/table.rs#L856-L866

Consequently that last no-op plan is added, but since it has wrong partitioning it causes the sanity check panic.

gruuya:
Closes #217.

### DIFF
--- a/datafusion_iceberg/src/table.rs
+++ b/datafusion_iceberg/src/table.rs
@@ -24,6 +24,7 @@ use crate::{
     statistics::manifest_statistics,
 };
 use datafusion::common::NullEquality;
+use datafusion::physical_plan::empty::EmptyExec;
 use datafusion::{
     arrow::datatypes::{DataType, Field, Schema as ArrowSchema, SchemaBuilder, SchemaRef},
     catalog::Session,
@@ -53,7 +54,6 @@ use datafusion::{
     scalar::ScalarValue,
     sql::parser::DFParserBuilder,
 };
-
 use iceberg_rust::spec::{schema::Schema, view_metadata::ViewRepresentation};
 use iceberg_rust::{
     arrow::write::write_parquet_partitioned, catalog::tabular::Tabular, error::Error,
@@ -830,7 +830,7 @@ async fn table_scan(
         .await?;
 
     // Create plan for partitions without delete files
-    let file_groups = data_file_groups
+    let file_groups: Vec<_> = data_file_groups
         .into_values()
         .map(|x| {
             x.into_iter()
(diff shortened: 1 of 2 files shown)


JanKaul:
Thanks for the PR! I can have a look at the error. Looks like something with Materialized Views.

gruuya:
> Thanks for the PR! I can have a look at the error. Looks like something with Materialized Views.

Oh thanks, just force-pushed a fix before seeing your message. 

It seems that sometimes there can be no plans and no data file groups either, so in that case it's useful to use the no-op plan.

gruuya:
That said it might be a bit more elegant to not rely on the no-op/empty parquet scan and instead be more explicit with
```rust
    match plans.len() {
        0 => {
            let projected_schema = projection
                .map(|p| arrow_schema.project(&p))
                .transpose()?
                .unwrap_or(arrow_schema.as_ref().clone());
            Ok(Arc::new(EmptyExec::new(Arc::new(projected_schema))))
        }
        1 => Ok(plans.remove(0)),
        _ => Ok(Arc::new(UnionExec::new(plans))),
    }
```
Let me know what you think about that.

JanKaul:
Yeah, I think that would be best. Thanks again!

Issue #212, opened 2025-07-16
Incorrect manifest list entries for (equality) delete files


gruuya:
## TLDR
Whilst the [`content`](https://github.com/apache/iceberg/blob/31daaedd13daf0e7ee4781fa5e768f9fe5ba5dc9/format/spec.md?plain=1#L656) field of `DataFile`s produced when writing equality deletes is set correctly (`2`) the corresponding [`content`](https://github.com/apache/iceberg/blob/31daaedd13daf0e7ee4781fa5e768f9fe5ba5dc9/format/spec.md?plain=1#L617) in `ManifestListEntry` isn't.

Instead of being set to `Deletes` (i.e. 1) it is hard-coded to `Data` (i.e. 0) https://github.com/JanKaul/iceberg-rust/blob/0e46fe3cc8cef254a23f197c1ba04b28556ac79c/iceberg-rust/src/table/manifest.rs#L239
This is fine when proper data files come in, but any follow-up deletes will end up in the same manifest list entry, thus having a misleading `content` value.

In turn this causes query engines such as Spark and DuckDB to treat those as data files instead of delete files, and thus append those deleted rows, as opposed to deduct them from the previous snapshot.

## Detailed investigation

Starting off of from the equality delete test
```diff
diff --git a/datafusion_iceberg/tests/equality_delete.rs b/datafusion_iceberg/tests/equality_delete.rs
index bcf1e9d..b395f2e 100644
--- a/datafusion_iceberg/tests/equality_delete.rs
+++ b/datafusion_iceberg/tests/equality_delete.rs
@@ -17,10 +17,11 @@ use iceberg_rust::{
 };
 use iceberg_sql_catalog::SqlCatalog;
 use std::sync::Arc;
+use object_store::local::LocalFileSystem;
 
 #[tokio::test]
 pub async fn test_equality_delete() {
-    let object_store = ObjectStoreBuilder::memory();
+    let object_store = ObjectStoreBuilder::Filesystem(Arc::new(LocalFileSystem::new()));
 
     let catalog: Arc<dyn Catalog> = Arc::new(
         SqlCatalog::new("sqlite://", "warehouse", object_store.clone())
@@ -79,7 +80,7 @@ pub async fn test_equality_delete() {
 
     let table = Table::builder()
         .with_name("orders")
-        .with_location("/test/orders")
+        .with_location("/Users/gruuya/test/orders")
         .with_schema(schema)
         .with_partition_spec(partition_spec)
[...]

PR #227, opened 2025-08-04
fix: write delete files to separate manifest list entries

gruuya:
- make sure manifest list entries are exclusively for data or delete content, and never mixed, since that will confuse other readers
- when executing the append operation try to find and re-use an existing manifest list entry with content that has the most files to add, instead of always appending to data entries
- extend the equality delete integration test, and verify DuckDB can read it correctly as well

Closes #212.

PR #228, opened 2025-08-05
use separate manifest files for delete files

JanKaul:
This PR implements an alternative approach to solve #212.

It selects two manifest files, one for data and one for delete files. The aim is to limit the number of additional manifest files that have to be created.

### DIFF
--- a/datafusion_iceberg/tests/equality_delete.rs
+++ b/datafusion_iceberg/tests/equality_delete.rs
@@ -1,5 +1,10 @@
-use datafusion::{arrow::error::ArrowError, assert_batches_eq, prelude::SessionContext};
+use datafusion::{
+    arrow::{error::ArrowError, record_batch::RecordBatch},
+    assert_batches_eq,
+    prelude::SessionContext,
+};
 use datafusion_iceberg::catalog::catalog::IcebergCatalog;
+use duckdb::Connection;
 use futures::stream;
 use iceberg_rust::catalog::identifier::Identifier;
 use iceberg_rust::catalog::tabular::Tabular;
@@ -16,11 +21,15 @@ use iceberg_rust::{
     table::Table,
 };
 use iceberg_sql_catalog::SqlCatalog;
+use object_store::local::LocalFileSystem;
 use std::sync::Arc;
+use tempfile::TempDir;
 
 #[tokio::test]
 pub async fn test_equality_delete() {
-    let object_store = ObjectStoreBuilder::memory();
+    let temp_dir = TempDir::new().unwrap();
+    let table_dir = format!("{}/test/orders", temp_dir.path().to_str().unwrap());
+    let object_store = ObjectStoreBuilder::Filesystem(Arc::new(LocalFileSystem::new()));
 
     let catalog: Arc<dyn Catalog> = Arc::new(
         SqlCatalog::new("sqlite://", "warehouse", object_store.clone())
@@ -79,7 +88,7 @@ pub async fn test_equality_delete() {
 
     let table = Table::builder()
         .with_name("orders")
-        .with_location("/test/orders")
+        .with_location(&table_dir)
         .with_schema(schema)
         .with_partition_spec(partition_spec)
         .build(&["test".to_owned()], catalog.clone())
@@ -194,4 +203,28 @@ pub async fn test_equality_delete() {
         "+----+-------------+------------+------------+--------+",
     ];
     assert_batches_eq!(expected, &batches);
+
+    let latest_version = table
+        .object_store()
+        .get(&object_store::path::Path::from(format!(
+            "{table_dir}/metadata/version-hint.text"
+        )))
+        .await
+        .unwrap()
+        .bytes()
+        .await
+        .unwrap();
+    let latest_version_str = std::str::from_utf8(&latest_version).unwrap();
+
+    let conn = Connection::open_in_memory().unwrap();
+    conn.execute("install iceberg", []).unwrap();
+    conn.execute("load iceberg", []).unwrap();
+
+    let duckdb_batches: Vec<RecordBatch> = conn
+        .prepare("select * from iceberg_scan(?) order by id")
+        .unwrap()
+        .query_arrow([latest_version_str])
+        .unwrap()
+        .collect();
+    assert_batches_eq!(expected, &duckdb_batches);
 }
--- a/iceberg-rust/src/table/manifest_list.rs
+++ b/iceberg-rust/src/table/manifest_list.rs
@@ -988,10 +988,18 @@ impl<'schema, 'metadata> ManifestListWriter<'schema, 'metadata> {
     }
 
     pub(crate) async fn finish(
-        self,
+        mut self,
         snapshot_id: i64,
         object_store: Arc<dyn ObjectStore>,
     ) -> Result<String, Error> {
+        if let Some(selected_data_manifest) = self.selected_data_manifest.take() {
+            self.writer.append_ser(selected_data_manifest)?;
+        }
+
+        if let Some(selected_delete_manifest) = self.selected_delete_manifest.take() {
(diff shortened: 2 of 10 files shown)


datafusion_iceberg/tests/equality_delete.rs
```diff
@@ -194,4 +203,28 @@ pub async fn test_equality_delete() {
...
+        .object_store()
+        .get(&object_store::path::Path::from(format!(
+            "{table_dir}/metadata/version-hint.text"
+        )))
+        .await
+        .unwrap()
+        .bytes()
+        .await
+        .unwrap();
+    let latest_version_str = std::str::from_utf8(&latest_version).unwrap();
```
gruuya, line 217: This was a bit unfortunate, since i realized that:
- just loading the table won't give you the full path to latest metadata file
- the `version-hint.text` (which does contain the full path) itself  is not readable by duckdb
   ```
   Iceberg metadata file not found for table version '/Users/gruuya/EDB/store/warehouse/tpch/query12/. metadata/00000-3f569e94-5601-48f3-9199-8d71df4ea7b0.metadata.json' using 'none' compression and format(s): 'v%s%s.metadata.json,%s%s.metadata.json'
   ```

There are ways to [customize](https://duckdb.org/docs/stable/core_extensions/iceberg/overview.html#working-with-alternative-metadata-naming-conventions) how duckdb parses the version to read, but i think those work only when the version file is not detected in the first place. Either way it's a separate issue.

iceberg-rust/src/table/transaction/append.rs
```diff
@@ -127,70 +129,122 @@ pub(crate) fn select_manifest_partitioned(
...
+                    continue;
+                };
+
+                if selected_row_count.is_some_and(|x| x > row_count) {
+                    manifest_list_writer.append_ser(selected_manifest)?;
+                    selected_delete_state = Some((Some(row_count), manifest));
+                    continue;
+                } else {
+                    manifest_list_writer.append_ser(manifest)?;
+                    continue;
```
gruuya, line 235: These `continue`s are redundant; likewise above (lines 143, 147, 161, 165, 211, 214).

Besides that I find the let-some-else-continue less readable so i refactored those in my pr, though that's not a big deal.

iceberg-rust/src/table/manifest_list.rs
```diff
@@ -664,14 +671,19 @@ impl<'schema, 'metadata> ManifestListWriter<'schema, 'metadata> {
     ///     object_store,
     /// ).await?;
     /// ```
-    pub(crate) async fn append_filtered_and_finish(
-        mut self,
+    pub(crate) async fn append_filtered(
```
gruuya, line 681: All these need docs updates that it doesn't `finish` on it's own now.

iceberg-rust/src/table/manifest_list.rs
```diff
@@ -395,7 +398,8 @@ impl<'schema, 'metadata> ManifestListWriter<'schema, 'metadata> {
         Ok(Self {
             table_metadata,
             writer,
-            selected_manifest: Some(manifest),
+            selected_data_manifest: Some(data_manifest),
+            selected_delete_manifest: delete_manifest,
```
gruuya, line 404: I like that there are both data and delete manifest entries selected. It's more explicit and better than in my pr where i only pick to continue in the one which has more new files to be added to.

iceberg-rust/src/table/transaction/append.rs
```diff
@@ -127,70 +129,122 @@ pub(crate) fn select_manifest_partitioned(
...
         }
     }
-    selected_state
-        .map(|(_, entry)| SelectedManifest {
-            manifest: entry,
-            file_count_all_entries,
-        })
-        .ok_or(Error::NotFound("Manifest for insert".to_owned()))
+    let (_, data_manifest) =
+        selected_data_state.ok_or(Error::NotFound("Manifest for insert".to_owned()))?;
```
gruuya, line 241: Maybe this shouldn't be an error but instead, the entire `Option<ManifestListEntry>` should be returned, since the caller (`ManifestListWriter::from_existing`) knows how to handle that correctly.

Granted I think this scenario can only happen if we have an empty/new table and we wont to append some deletes to it, or some such edge case.

gruuya:
Btw those two failing mat-view tests are solved by this https://github.com/JanKaul/iceberg-rust/pull/227/commits/0dc313a2e273cf155cafc7b0940be846a6c5c7ea (otherwise you end up generating identical data and delete manifest paths).

datafusion_iceberg/tests/equality_delete.rs
JanKaul: Hmm, is the naming of the metadata.json wrong then?

iceberg-rust/src/table/transaction/append.rs
JanKaul: I'll have look, thanks. I wanted to quickly sketch out the general idea. Haven't looked into that yet.

JanKaul:
> Btw those two failing mat-view tests are solved by this [0dc313a](https://github.com/JanKaul/iceberg-rust/commit/0dc313a2e273cf155cafc7b0940be846a6c5c7ea) (otherwise you end up generating identical data and delete manifest paths).

Awesome, thanks! I was stuck on how to solve that one.

datafusion_iceberg/tests/equality_delete.rs
gruuya: I think the naming is fine, it's just that version-hint.text should contain `00000-3f569e94-5601-48f3-9199-8d71df4ea7b0` (e.g. for a file named `00000-3f569e94-5601-48f3-9199-8d71df4ea7b0.metadata.json`), so that it complies with the format `%s%s.metadata.json` (the other %s is for the compression codec.

I also know spark doesn't expect that format either
```python
In [30]: df = spark.read.format("iceberg").load("/Users/gruuya/EDB/store/test/orders")
25/07/16 09:12:42 WARN HadoopTableOperations: Error reading version hint file /Users/gruuya/EDB/store/test/orders/metadata/version-hint.text
java.lang.NumberFormatException: For input string: "/Users/gruuya/EDB/store/test/orders/metadata/v2.metadata.json"
```
but I'm not sure exactly what's the proper format for it then.

iceberg-rust/src/table/transaction/operation.rs
```diff
@@ -138,40 +138,82 @@ impl Operation {
...
-                    manifest_list_writer
-                        .append_multiple_and_finish(
-                            new_datafile_iter,
-                            snapshot_id,
-                            n_splits,
-                            object_store,
-                        )
-                        .await?
-                };
+                if n_data_files != 0 {
```
gruuya: nit: could use something like  
```rust
for (content, files) in [(Content::Data, data_files), (Content::Deletes, delete_files)]
```
to deduplicate some code below.

gruuya (approved):

gruuya:
Fwiw, I've also been seeing the [no-disk-space](https://github.com/JanKaul/iceberg-rust/actions/runs/16770800255/job/47485232178?pr=228#step:6:207) error during trino image pull.

I've done some diagnosing on it and it seems like `/usr/local` is inflated, and just by removing `/usr/local/lib/android` around 10GB could be [reclaimed](https://github.com/JanKaul/iceberg-rust/actions/runs/16745871619/job/47404175959#step:6:56).

iceberg-rust/src/table/transaction/append.rs
JanKaul: I think think there is no valid scenario where there shouldn't be a data manifest selected here. And I think not having an `Option` in `SelectedManifest` is the most explicit way to express that. This way an Error has to be thrown, which in my opinion is  the correct behavior. If you pass the Option to the caller, you rely on correct handling.

PR #229, opened 2025-08-06
Multiple data sequence numbers per commit

### LINKED ISSUE #222: Multiple data sequence numbers in one commit
We have a use case where we generate N changesets. Each changeset is a (data_files, delete_files) tuple. The changesets should have distinct data sequence numbers so that delete files from later changesets apply to data files from earlier changesets. However, we want to commit all changesets together in one Iceberg commit.

Is there any to do this today in iceberg-rust? Can you think of any interface changes to support this? I was thinking of something like this:

```
let mut tx = table.new_transaction(None);
tx = tx.append_delete(delete1);
tx = tx.append_data(data1);
tx = tx.new_data_sequence_number();
tx = tx.append_delete(delete2);
tx = tx.append_data(data2);
tx = tx.new_data_sequence_number();
// ...
tx = tx.append_delete(deleteN);
tx = tx.append_data(dataN);
tx.commit().await;
```

Or maybe like this:
```
let mut tx = table.new_transaction(None);
tx = tx.append_delete_with_dsn_increment(delete1, 1);
tx = tx.append_data_with_dsn_increment(data1, 1);
tx = tx.append_delete_with_dsn_increment(delete2, 2);
tx = tx.append_data_with_dsn_increment(data2, 2);
// ...
tx = tx.append_delete_with_dsn_increment(deleteN, N);
tx = tx.append_data_with_dsn_increment(dataN, N);
tx.commit().await;
```

What do you think?

### LINKED ISSUE #232: Deletes passed to `TableTransaction.append_delete()` get silently lost
## Steps to reproduce
```
let mut tx = table.new_transaction(None);
tx = tx.append_data(data_file);
tx = tx.append_delete(delete_file);
tx.commit().await;
```

## Observed behavior
- The commit does not include `delete_file`

## Expected behavior
- The commit should include `delete_file`. Alternatively if such a transaction is not supported then the `tx.append_*` call should error out.

## Notes
- if `.append_delete` is reordered before `.append_data` then everything works as expected
- if `.append_delete` is on its own without `.append_data` then everything works as expected

SergeiPatiakin:
- closes https://github.com/JanKaul/iceberg-rust/issues/222
- fixes https://github.com/JanKaul/iceberg-rust/issues/232
- introduce the concept of Data Sequence Number Group (DSN group), defined as the set of DataFiles with the same data sequence number
- allow a commit to consist of multiple DSN groups
- allow consumers to call `transaction.new_data_sequence_number()` to start a new DSN group

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -20,7 +20,7 @@ jobs:
         android: true
         dotnet: true
         haskell: true
-        large-packages: false
+        large-packages: true
         docker-images: false
         swap-storage: false
     - uses: actions/checkout@v3
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -13,6 +13,16 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
+    - name: Free disk space
+      uses: jlumbroso/free-disk-space@main
+      with:
+        tool-cache: false
+        android: true
+        dotnet: true
+        haskell: true
+        large-packages: false
+        docker-images: false
+        swap-storage: false
     - uses: actions/checkout@v3
     - name: Build
       run: cargo build --verbose
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -13,6 +13,16 @@ jobs:
   test:
     runs-on: ubuntu-latest
     steps:
+    - name: Free disk space
+      uses: jlumbroso/free-disk-space@main
+      with:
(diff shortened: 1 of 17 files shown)


JanKaul:
Thanks for the PR. I have my doubts about this approach: https://github.com/JanKaul/iceberg-rust/issues/222#issuecomment-3173845797

SergeiPatiakin:
Closing in favor of https://github.com/JanKaul/iceberg-rust/pull/251

PR #234, opened 2025-08-07
fix: make version-hint.text compatible with other readers

gruuya:
Instead of storing the full path to the latest metadata file store only the relevant version string.

1. in case of the [file-system format](https://iceberg.apache.org/spec/#file-system-tables) that's just a single number `V` from `v<V>.metadata.json`. This works for both Spark and DuckDB.
2. in case of the [metastore](https://iceberg.apache.org/spec/#metastore-tables) format this is the number + the uuid, e.g. `00002-2ece07c9-4321-4926-8f05-f7eedce4d6d0` in `00002-2ece07c9-4321-4926-8f05-f7eedce4d6d0.metadata.json`, which DuckDb will recognize by default. 

Spark on the other hand doesn't accept that second format
```python
In [3]: df = spark.read.format("iceberg").load("/Users/gruuya/warehouse/test/orders")
25/08/07 12:34:43 WARN HadoopTableOperations: Error reading version hint file /Users/gruuya/warehouse/test/orders/metadata/version-hint.text
java.lang.NumberFormatException: For input string: "00002-2ece07c9-4321-4926-8f05-f7eedce4d6d0"
```
Even if the uuid part is trimmed I see ` org.apache.iceberg.exceptions.ValidationException: Metadata file for version 2 is missing`, suggesting Spark only tries to parse the provided hint as a number to be interpolated into the file-system format.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2756,6 +2756,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f90f7dce0722e95104fcb095585910c0977252f286e354b5e3bd38902cd99988"
 
 [[package]]
+name = "futures-timer"
+version = "3.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f288b0a4f20f9a56b5d1da57e2227c661b7b16168e2f72365f57b63326e29b24"
+
+[[package]]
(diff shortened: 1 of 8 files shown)


JanKaul:
Thanks a lot for looking into this.

PR #239, opened 2025-08-11
fix: conflict when projecting a field not present in equality deletes

### LINKED ISSUE #238: Panic when projecting columns not present in the equality deletes
The current code doesn't handle well the case when a projection selects only column(s) not present in equality ids, nor in general potential differences in equality ids across different deletion files.

For instance when adding the following code at the end of the `test_equality_delete` test
```rust 
    // Test that projecting a column that is not included in equality deletes works
    ctx.sql(
        "INSERT INTO warehouse.test.orders (id, customer_id, product_id, date, amount) VALUES
                (7, 3, 2, '2020-01-01', 2),
                (8, 2, 1, '2020-02-02', 3),
                (9, 1, 3, '2020-01-01', 1);",
    )
        .await
        .expect("Failed to create query plan for insert")
        .collect()
        .await
        .expect("Failed to insert values into table");

    let batches = ctx
        .sql("select sum(amount) from warehouse.test.orders")
        .await
        .expect("Failed to create plan for select")
        .collect()
        .await
        .expect("Failed to execute select query");

    let expected = [
        "+-----------------------------------+",
        "| sum(warehouse.test.orders.amount) |",
        "+-----------------------------------+",
        "| 13                                |",
        "+-----------------------------------+",
    ];
    assert_batches_eq!(expected, &batches);
```

leads to a panic
```
thread 'test_equality_delete' panicked at /Users/gruuya/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/arrow-schema-55.2.0/src/schema.rs:382:10:
index out of bounds: the len is 1 but the index is 1
```

gruuya:
Closes #238.

### DIFF
--- a/datafusion_iceberg/src/table.rs
+++ b/datafusion_iceberg/src/table.rs
@@ -408,30 +408,22 @@ async fn table_scan(
 
     let file_schema: SchemaRef = Arc::new((schema.fields()).try_into().unwrap());
 
-    let projection = projection.cloned().or_else(|| {
-        Some(
-            arrow_schema
-                .fields()
-                .iter()
-                .enumerate()
-                .map(|(i, _)| i)
-                .collect(),
-        )
-    });
+    // If no projection was specified default to projecting all the fields
+    let projection = projection
+        .cloned()
+        .unwrap_or((0..arrow_schema.fields().len()).collect_vec());
 
-    let projection_expr: Option<Vec<_>> = projection.as_ref().map(|projection| {
-        projection
-            .iter()
-            .enumerate()
-            .map(|(i, id)| {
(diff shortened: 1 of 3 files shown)


datafusion_iceberg/src/table.rs
```diff
@@ -408,30 +408,22 @@ async fn table_scan(
 
     let file_schema: SchemaRef = Arc::new((schema.fields()).try_into().unwrap());
 
-    let projection = projection.cloned().or_else(|| {
```
gruuya, line 411: This was always `Some`, so might as well make it explicit.

JanKaul:
Thanks for finding and fixing this. The solution looks much better.

PR #244, opened 2025-08-20
fix: update `last-updated-ms` field when updating a table

gruuya:
Without this some catalogs (file, glue, s3tables, sql) will retain the value that was entered during table creation, since the entire metadata struct is fetched from the cache.

This in turn can make spark error out on validation when trying to read the table:
```
IllegalArgumentException: Invalid update timestamp 1755628214467: before last snapshot log entry at 1755628701901
```

### DIFF
--- a/iceberg-rust/src/catalog/commit.rs
+++ b/iceberg-rust/src/catalog/commit.rs
@@ -9,8 +9,6 @@
 //! All changes are made atomically - either all updates succeed or none are applied.
 //! Requirements are checked first to ensure concurrent modifications don't corrupt state.
 
-use std::collections::HashMap;
-
 use iceberg_rust_spec::{
     spec::{
         partition::PartitionSpec,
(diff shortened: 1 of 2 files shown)


iceberg-rust/src/catalog/commit.rs
```diff
@@ -530,6 +530,12 @@ pub fn apply_table_updates(
             }
         };
     }
+
+    // Lastly make sure `last-updated-ms` field is up-to-date
+    metadata.last_updated_ms = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .unwrap()
+        .as_millis() as i64;
```
gruuya, line 538: I think this construct should be extracted to some util function and re-used in a couple of places (TableMetadata, Snapshot, here, etc.).

Any idea what is the best place to do so?

JanKaul:
Great, thanks for fixing the bug!

PR #245, opened 2025-08-26
fix: use correct number of splits for deletes

gruuya:
Currently we're erroneously using the count of data splits (`n_data_splits`) even for deletes. 

Besides being generally incorrect, in situations where we have zero data splits but non-zero delete splits, we'll pass the zero down to `append_multiple`, which will lead to a `attempt to subtract with overflow` panic [here](https://github.com/JanKaul/iceberg-rust/blob/db8fd256deba9b1969dc77c9699fbea27382dabd/iceberg-rust/src/table/transaction/append.rs#L93).

### DIFF
--- a/iceberg-rust/src/table/transaction/operation.rs
+++ b/iceberg-rust/src/table/transaction/operation.rs
@@ -139,10 +139,6 @@ impl Operation {
                     )?
                 };
 
-                let n_data_splits = manifest_list_writer.n_splits(n_data_files, Content::Data);
-                let n_delete_splits =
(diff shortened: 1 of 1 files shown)


gruuya:
I see that `test_insert_csv` is failing, but it's unrelated to this pr, as it's also failing on main apparently.

gruuya:
> I see that test_insert_csv is failing

Btw i found the issue, will open a separate pr for it.

EDIT: here it is https://github.com/JanKaul/iceberg-rust/pull/248

JanKaul:
Thanks for the catch and fix!

PR #247, opened 2025-08-26
use filesystem objectstore

JanKaul:
Fixes the insert_csv_test


gruuya:
While this does fix the test it doesn't address the root cause, which is that the fake object store url in case of relative paths is usually unwrapped to `file://`, thus overwriting the default FS object store with a random one.

One way to fix this root cause is outlined in https://github.com/JanKaul/iceberg-rust/pull/248

PR #248, opened 2025-08-26
fix: make fake object store url always parsable

gruuya:
Since previously `fake_object_store_url` returned an `Option`, the natural instinct was to use `ObjectStoreUrl::local_filesystem()` as default when unwrapping it.

However this meant that the default `file://` object store in the registry gets overwritten with a random unrelated one for relative table paths. This was the case for `test_insert_csv`, where in-memory object store replaced the default file one, hence the test [failed](https://github.com/JanKaul/iceberg-rust/actions/runs/17126323542/job/48578786198#step:7:193) with `Failed to execute query plan.: Shared(ObjectStore(NotFound { path: "home/runner/work/iceberg-rust/iceberg-rust/datafusion_iceberg/testdata/tpch/lineitem.csv", source: NoDataInMemory { path: "home/runner/work/iceberg-rust/iceberg-rust/datafusion_iceberg/testdata/tpch/lineitem.csv" } }))`

Instead of returning `Option`, this PR makes `fake_object_store_url` always return a `ObjectStoreUrl`, which is a fake url anyway, so url validation should be handled by other components.

In addition it tries to retain the human-readable segments (paths), while using quasi-url-encoding to escape the slash and column characters (`/` -> `-2F`, and `:` -> `-3A`)

### DIFF
--- a/catalogs/iceberg-file-catalog/src/lib.rs
+++ b/catalogs/iceberg-file-catalog/src/lib.rs
@@ -896,10 +896,7 @@ pub mod tests {
             .await
             .unwrap();
 
-        assert_eq!(
-            std::str::from_utf8(&version_hint).unwrap(),
-            "s3://warehouse/tpch/lineitem/metadata/v1.metadata.json"
-        );
+        assert_eq!(std::str::from_utf8(&version_hint).unwrap(), "1");
 
         let files = object_store.list(None).collect::<Vec<_>>().await;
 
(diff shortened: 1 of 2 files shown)


gruuya:
Fwiw, this is also aligned with how it's done in delta-rs https://github.com/delta-io/delta-rs/blob/1f0b4d0965a85400c1effc6e9b4c7ebbb6795978/crates/core/src/logstore/mod.rs#L515-L521

JanKaul:
Thanks for working on this! LGTM

PR #250, opened 2025-08-27
docs: default compression level is 3

satlank:
Apologies for the anemic change here, but I noticed the misalignment between comment and code while reading through the codebase.


JanKaul:
Thanks for spotting the wrong documentation.

PR #251, opened 2025-08-27
AppendSequenceGroups operation

SergeiPatiakin:
This PR follows the suggestion https://github.com/JanKaul/iceberg-rust/issues/222#issuecomment-3179443665 to add a new AppendSequenceGroups operation. This allows consumers to use multiple sequence groups per commit. Delete files from later sequence groups can be applied to data files from earlier sequence groups.


JanKaul:
Thanks for the contribution. Looks good to me.

PR #263, opened 2025-09-29
Add some bucket transformation

pierre-marijon:
Hello,

First of all, thank you for your work.

In my use case, some transformations were missing, so I added them.

To implement bucket transformation, I added a `fasthash` dependency. This dependency introduces more functionality than necessary. If this is a problem, I can look for another dependency or simply replace it with a local implementation of murmur3. (Maybe I should move declaration of dependency from main Cargo.toml to `iceberg-rust/Cargo.toml`)

I added a test to verify that the `fasthash::murmur3::hash32_with_seed` implementation matches the [Apache Iceberg specification](https://iceberg.apache.org/spec/#appendix-b-32-bit-hash-requirements).

If you would prefer me to add other transformations or if you would like more test values, please let me know and I can improve my code.

Best

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3331,6 +3331,7 @@ dependencies = [
  "itertools 0.14.0",
  "lazy_static",
  "lru",
+ "murmur3",
  "object_store",
  "parquet",
  "pin-project-lite",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ itertools = "0.14.0"
 lazy_static = "1.5.0"
 lru = "0.16.0"
 object_store = { version = "0.12", features = ["aws", "gcp"] }
+fasthash = { version = "0.4" }
 parquet = { version = "56", features = ["async", "object_store"] }
 pin-project-lite = "0.2"
 regex = "1.11.1"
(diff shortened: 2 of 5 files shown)


JanKaul:
This is great! Thanks a lot for your work. One nit pick: I'm already using the `murmur1` library in the `iceberg-rust-spec` crate: https://github.com/JanKaul/iceberg-rust/blob/main/iceberg-rust-spec/Cargo.toml#L19

I would like to only use one library for the murmur hash. I actually don't have a strong opinion on which one. But if you could only use either of the libraries, that would be great.

pierre-marijon:
Which crate use or not use is also my main concern.

So I just made a pro-cons list with my point of view.

fasthash pro:
- simple API

fasthash cons:
- C dependency
- large amount of not interesting code (so huge compilation time)

murmur3 pro:
- pure rust
- just what we need

murmur3 cons:
- little bit annoying api
- usage of `std::io::Read` API force return of an `std::io::Result` but it's seems very improbable in our use case.

Finally murmur3 have less clean API but pure Rust and smaller compilation time seems better for me.

JanKaul:
Thanks a lot for fixing it so quickly. LGTM!

PR #270, opened 2025-11-03
fix: a substrack overflow emit in debug mode

pierre-marijon:
I just spot a simple annoying bug which only appeared in debug mode. Use `abs_diff` instead of classic `diff` seems solve this bug in debug mode without impact on other test.


JanKaul:
Thanks a lot for fixing this!

PR #271, opened 2025-11-03
Add support for Microsoft Azure Storage Blob in ObjectStore

lizardoluis:
Adds support for Microsoft Azure Blob Storage to the object store abstraction in the project. It introduces Azure as storage backend, updates configuration parsing and builder logic, and provides thorough tests for Azure bucket detection and configuration. The changes ensure Azure can be used alongside AWS S3 and Google Cloud Storage with similar APIs and configuration patterns.

### DIFF
--- a/iceberg-rust/src/object_store/mod.rs
+++ b/iceberg-rust/src/object_store/mod.rs
@@ -6,6 +6,7 @@ use std::{fmt::Display, path::Path, str::FromStr, sync::Arc};
 
 use object_store::{
     aws::{AmazonS3Builder, AmazonS3ConfigKey, S3CopyIfNotExists},
+    azure::{AzureConfigKey, MicrosoftAzureBuilder},
     gcp::{GoogleCloudStorageBuilder, GoogleConfigKey},
     local::LocalFileSystem,
     memory::InMemory,
@@ -24,6 +25,8 @@ pub enum Bucket<'s> {
     S3(&'s str),
     /// GCS bucket
     GCS(&'s str),
+    /// Azure container
+    Azure(&'s str),
     /// No bucket
     Local,
 }
@@ -33,13 +36,14 @@ impl Display for Bucket<'_> {
         match self {
             Bucket::S3(s) => write!(f, "s3://{s}"),
             Bucket::GCS(s) => write!(f, "gs://{s}"),
+            Bucket::Azure(s) => write!(f, "https://{s}"),
             Bucket::Local => write!(f, ""),
         }
     }
 }
 
 impl Bucket<'_> {
-    /// Get the bucket and coud provider from the location string
+    /// Get the bucket and cloud provider from the location string
     pub fn from_path(path: &str) -> Result<Bucket<'_>, Error> {
         if path.starts_with("s3://") || path.starts_with("s3a://") {
             let prefix = if path.starts_with("s3://") {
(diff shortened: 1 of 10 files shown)


iceberg-rust/src/object_store/mod.rs
```diff
@@ -125,6 +151,9 @@ impl ObjectStoreBuilder {
     /// Set config value for builder
     pub fn with_config(self, key: ConfigKey, value: impl Into<String>) -> Self {
         match (self, key) {
+            (ObjectStoreBuilder::Azure(azure), ConfigKey::Azure(key)) => {
+                ObjectStoreBuilder::Azure(Box::new(azure.with_config(key, value)))
+            }
             (ObjectStoreBuilder::S3(aws), ConfigKey::AWS(key)) => {
                 ObjectStoreBuilder::S3(Box::new(aws.with_config(key, value)))
             }
```
JanKaul: I realized that it's not good to parse the config string into a `ConfigKey` before passing it to the builder, because there is no knowledge about which provider(s3,gcs,azure) to parse it for. This is an issue if the providers have matching config keys. That's why one of the tests is failing.
Instead we need to pass the string to `with_config()` and then parse the `ConfigKey` inside of it when we have the appropriate information.
I'll create a PR to fix this.

JanKaul: https://github.com/splitgraph/iceberg-rust/pull/20

lizardoluis: In our system, we use the `ObjectStoreScheme::parse()` with the URL to define which builder to use.

JanKaul: Do you use the `with_config` method? Or do you build the entire Builder beforehand and then pass it to `ObjectStoreBuilder`?

iceberg-rust/src/object_store/mod.rs
```diff
@@ -93,6 +112,9 @@ pub enum ConfigKey {
 impl FromStr for ConfigKey {
     type Err = object_store::Error;
     fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if let Ok(x) = s.parse() {
+            return Ok(ConfigKey::Azure(x));
+        };
         if let Ok(x) = s.parse() {
             return Ok(ConfigKey::AWS(x));
         };
```
JanKaul, line 120: Could you swap the order of Azure and AWS? Because then the tests shouldn't fail anymore.

This however could lead to problems on your side. That's why I'm proposing to update the `with_config` method. Which should fix the issues in either case.

The problem here is that the output ConfigKey depends on the order of the which the providers are parsed. A keywork like "endpoint" which is supported by multiple providers, will always be parsed as a key of the provider that is listed here first. Even though it might want to be used for a different provider.

lizardoluis: I can do. Should I pick your changes on the other PR? Or are you going to merge them? 
I also found a problem with the way I'm extracting the buckets. I'll fix that first and set this PR as draft until then

lizardoluis: Ah, just noticed that your PR is on our fork. I'll merge it

JanKaul: I hope it doesn't cause any problems on your side but I would like to inlcude my PR. This should make it more robust when using config string with different providers.

When my PR is merged, you also don't have to swap the order, as this is handled automatically.

iceberg-rust/src/object_store/mod.rs
lizardoluis: Both ways. We have some fixed options that we pass to the builder in a structured way, but the user can set more and those we add with `with_config()`.

JanKaul:
Looks good to me now

lizardoluis:
Ok, I didn’t add any other changes — it can be merged as is. Thanks for handling your side of the fix.

I marked it as a draft because I wasn’t sure about the [bucket parser](https://github.com/JanKaul/iceberg-rust/pull/271/files#diff-3960679c2064ce8c0ccc46a469f25ccf502a37dfa099cfb49bb6cd21347e392aR70-R80).

Azure uses a different URL format (`https://myaccount.blob.core.windows.net/mycontainer/myblob`), so I wasn’t sure whether the parser should return just `mycontainer` (as I implemented) or the full path `myaccount.blob.core.windows.net/mycontainer`.

JanKaul:
I don't have a lot of experience with Azure. Well, we can just fix it in another PR if the need arises.

Thanks for the PR!

PR #272, opened 2025-11-19
Set default for missing key_metadata field

osipovartem:
If there is no optional column in the avro snap file, then without a default, there will be an error during deserialization.


JanKaul:
Thanks for fixing this!

PR #275, opened 2025-11-25
Enhance Azure URI parser

lizardoluis:
Azure has a quite complex URIs format, when compared to AWS and GCP. With these changes, I'm bold trying to cover all the formats.

### DIFF
--- a/iceberg-rust/src/object_store/mod.rs
+++ b/iceberg-rust/src/object_store/mod.rs
@@ -18,6 +18,24 @@ use crate::error::Error;
 pub mod parse;
 pub mod store;
 
+/// Azure endpoint types
+#[derive(Debug)]
+pub enum AzureEndpointType {
+    /// Distributed File System
+    DFS,
+    /// Binary Large Object
+    Blob,
+}
+
+impl Display for AzureEndpointType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            AzureEndpointType::DFS => write!(f, "dfs"),
+            AzureEndpointType::Blob => write!(f, "blob"),
+        }
+    }
+}
+
 /// Type for buckets for different cloud providers
 #[derive(Debug)]
 pub enum Bucket<'s> {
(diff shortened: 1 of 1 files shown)


iceberg-rust/src/object_store/mod.rs
```diff
@@ -36,7 +41,9 @@ impl Display for Bucket<'_> {
         match self {
             Bucket::S3(s) => write!(f, "s3://{s}"),
             Bucket::GCS(s) => write!(f, "gs://{s}"),
-            Bucket::Azure(s) => write!(f, "https://{s}"),
+            Bucket::Azure { account, container } => {
+                write!(f, "abfs://{container}@{account}.dfs.core.windows.net/")
```
lizardoluis: I'm still checking if we should use abfs or https here

JanKaul:
I'm wondering whether we could reuse [`strip_prefix()`](https://github.com/JanKaul/iceberg-rust/blob/main/iceberg-rust-spec/src/util.rs#L20) here.

lizardoluis:
Setting as a draft until I can get a confirmation of these changes

lizardoluis:
Reopened it again for review please. 

The `strip_prefix` removes the bucket as well: ```assert_eq!(strip_prefix("s3://bucket/path"), "/path");```. And what we want there is the bucket.

lizardoluis:
With my latest changes https://github.com/JanKaul/iceberg-rust/pull/275/commits/349dc5fcf45e62eaa7214400d727866198854c92 I made it more robust by supporting both blob storage and adls.  Was the last one

JanKaul:
Thanks a lot! Is it ready to be merged now?

lizardoluis:
it is! You can merge please!

iceberg-rust/src/object_store/mod.rs
```diff
@@ -45,39 +79,100 @@ impl Display for Bucket<'_> {
...
-                "gs://"
-            };
-            path.trim_start_matches(prefix)
+            let prefix = extract_prefix(path);
+            path.trim_start_matches(prefix.as_str())
                 .split('/')
                 .next()
                 .map(Bucket::GCS)
                 .ok_or(Error::NotFound(format!("Bucket in path {path}")))
+        } else if path.starts_with("az://")
```
drachum, line 103: @lizardoluis have you considered bailing early to avoid the nesting chain? xD

PR #278, opened 2025-12-01
fix: Change Azure parquet URI to `abfss://`

lizardoluis:
@JanKaul One more change. Probably last one

With this, Spark connect was able to read the tables.


JanKaul:
Thanks!

Issue #155, opened 2025-04-07
Can't use AWS S3Tables with iceberg-rest-catalog


mildbyte:
I'm trying to configure the REST catalog as follows to read an AWS S3 Tables catalog:

```rust
use std::sync::Arc;

use aws_config::Region;
use iceberg_rust::{
    catalog::Catalog, object_store::ObjectStoreBuilder, spec::namespace::Namespace,
};

use iceberg_rest_catalog::apis::configuration::{AWSv4Key, Configuration};

async fn try_rest_catalog() {
    let configuration = Configuration {
        base_path: "https://s3tables.eu-west-2.amazonaws.com/iceberg".to_string(),
        aws_v4_key: Some(AWSv4Key {
            access_key: "ASIA-REDACTED".to_string(),
            secret_key: "REDACTED"
                .to_string()
                .into(),
            security_token: Some("REDACTED".to_string().into()),
            region: "eu-west-2".to_string(),
            service: "s3tables".to_string(),
        }),
        ..Configuration::default()
    };

    let builder = ObjectStoreBuilder::s3();

    let catalog = iceberg_rest_catalog::catalog::RestCatalog::new(
        Some("arn:aws:s3tables:eu-west-2:REDACTED:bucket/REDACTED"),
        configuration,
        builder,
    );

    let namespaces = catalog.list_namespaces(None).await.unwrap();
    for namespace in namespaces.iter() {
        println!("found namespace {}", namespace);
    }

    let namespace_to_list = Namespace::try_new(&vec!["public".to_string()]).unwrap();
    let tables = catalog.list_tabulars(&namespace_to_list).await.unwrap();

    for table in tables.iter() {
        println!("table {}", table.name())
    }
}

#[tokio::main]
async fn main() {
    tracing_subscriber::fmt::init();
    try_rest_catalog().await;
}
```

The security token is my session token that I got via `aws configure export-credentials` - without it, I was getting this error:

```
thread 'main' panicked at src/main.rs:64:58:
called `Result::unwrap()` on an `Err` value: InvalidFormat("Response status: 403 Forbidden, Response content: {\"message\":\"The security token included in the request is invalid.\"}")
```

I also applied this patch to iceberg-rest-catalog to get it to send the security token to AWS:

```diff
diff --git a/catalogs/iceberg-rest-catalog/src/apis/configuration.rs b/catalogs/iceberg-rest-catalog/src/apis/configuration.rs
index 003e506..a29b4b1 100644
[...]

PR #279, opened 2025-12-02
Fix empty request body serde

osipovartem:
Related to https://github.com/JanKaul/iceberg-rust/issues/155

`&serde_json::to_string(&request)` returns `"null"` for empty &() request. We sing the request with `"null"` body
and then
`let resp = client.execute(req).await?;`
 returns incorrect signature error


JanKaul:
Great, thanks for fixing this.

I'm guessing you want to use it with Datafusion version 50, right?

Could you make a PR against the df50 branch?

I recently updated the Datafusion version

PR #282, opened 2025-12-03
Write object_store to the cache when creating table

### LINKED ISSUE #280: RestCatalog Object store for table not found
https://github.com/JanKaul/iceberg-rust/blob/df50/catalogs/iceberg-rest-catalog/src/catalog.rs#L350-L363
When we call create_table we don't save object_store to the cache. 
And when trying to insert some data we call update_table and recieve `Object store for table not found` error since the cache is empty
```rust
let Some(object_store) = self.cache.read().unwrap().get(&identifier).cloned() else {
            return Err(Error::NotFound(format!(
                "Object store for table {}",
                &identifier
            )));
        };
```

osipovartem:
Closes #280

### DIFF
--- a/catalogs/iceberg-rest-catalog/src/catalog.rs
+++ b/catalogs/iceberg-rest-catalog/src/catalog.rs
@@ -359,6 +359,11 @@ impl Catalog for RestCatalog {
                     })
             })?;
 
+        self.cache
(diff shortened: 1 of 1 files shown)


JanKaul:
Thanks

Issue #291, opened 2025-12-18
Overwrite operation doesn't take into account filtered data files


osipovartem:
When we send Overwrite operation we can define a list of files to filter.
But during manifest list preparation we don't adjust ManifestListEnrty by filtered data files.

We need to take into account filtered data files in manifestListEntry and snapshots summary
Possible solution here

## Summary
- recompute manifest entries when reusing manifests, tracking removed file statistics for filtered overwrites
- propagate removal stats through manifest list filtering so overwrite manifest lists match the files actually written
- subtract overwritten files from snapshot summaries to keep totals consistent during merge/overwrite operations
https://github.com/Embucket/iceberg-rust/pull/45/files

PR #292, opened 2025-12-19
Take into account filtered files stats for Overwrite operation

osipovartem:
Closes https://github.com/JanKaul/iceberg-rust/issues/291


JanKaul:
I think these changes should be included: https://github.com/Embucket/iceberg-rust/pull/47.

Also fixes the test issues.

PR #293, opened 2025-12-22
Set corrent existing rows count

osipovartem:
When we filter data files we should also adjust added and existing rows and files count.
Since we use filtered stats only for counting removed data files and rows.
Example,
We have 
```
added_files_count: 1
added_rows_count: 6001215
```
**for the case with whole file overwrite from_existing_with_filter returns**
```
  added_files_count: 0
  existing_files_count: 1
  added_rows_count: 6001215
```
**And then we call append new data file with applying filtered stats**
```
  added_files_count: 1
  existing_files_count: 1
  deleted_files_count: 1
  added_rows_count: 12002430
```
Related to https://github.com/JanKaul/iceberg-rust/issues/291


osipovartem:
cc @JanKaul 
the same as https://github.com/Embucket/iceberg-rust/pull/48

JanKaul:
Thanks for the fix

PR #297, opened 2026-01-08
Add new `write_parquet_partitioned_opts`

arnaudbriche:
This PR add supports for configuring some previously static settings of the Parquet writing path.

### DIFF
--- a/iceberg-rust/src/arrow/write.rs
+++ b/iceberg-rust/src/arrow/write.rs
@@ -32,6 +32,7 @@
 //! # }
 //! ```
 
+use derive_builder::Builder;
 use futures::{
     channel::mpsc::{channel, Receiver, Sender},
     SinkExt, StreamExt, TryStreamExt,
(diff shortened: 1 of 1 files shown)


JanKaul:
Thanks for making these improvements.

My plan was to move to the [write_parquet_files](https://github.com/JanKaul/iceberg-rust/blob/main/datafusion_iceberg/src/table.rs#L1175) function in the future. It reuses the Datafusion ParquetSink and thus adheres to the Datafusion configuration parameters: https://datafusion.apache.org/user-guide/configs.html.
If you are using Datafusion you will likely be able configure the writer as is.

arnaudbriche:
Looks good, I'll try to switch.
Does it respect table partitioning as `write_parquet_partitioned` does ?

JanKaul:
Yes, and it performs much better if your table is not partitioned as Datafusion has a nice parallel parquet file writer.

arnaudbriche:
Cool. 
Not directly related to iceberg-rust, but as you seem very knowledgeable about Datafusion: I need my Parquet files to be sorted to optimize read perf, is there any setting in Datafusion sink that can guarantee Parquet rows are sorted ?

JanKaul:
I don't think so. Sorting files when running compaction should be part of this crate. But I didn't get around to implementing it.

arnaudbriche:
Switching to `write_parquet_data_files` works perfectly.
Setting `maximum_parallel_row_group_writers` to 1 and sorting upstream seems to produce sorted files.

PR #298, opened 2026-01-09
fix empty manifest list entries when using Overwrite

arnaudbriche:
While using Overwrite to implement partial table compaction, I've been confronted with a case where the resulting manifest list contained lot of "empty" manifest entries (pointing to no data files).
This small fix seems to discard these empty entries.


JanKaul:
Awesome, thanks for looking into this. The fix is great. However, I want to have a quick look and figure out where these empty files are coming from. As they shouldn't be produced in the first place. Do you by any chance have an example where I could reproduce this in a test?

arnaudbriche:
I do have a local dataset, but it's rather hard to transfer.
I investigated and these "empty" manifest entries are emitted when one of the manifest contains only "overwrites" files (all the data files in the manifest are filtered out)

PR #299, opened 2026-01-12
Add a check to ensure overwrites manifest actually exists in the snapshot on Overwrite operation

arnaudbriche:
When I first tried to use the Overwrite operation while implementing a custom table compaction engine, I used it in a wrong way but just using filename and not path, for overwrites.
The operation succedded but obviously the outcome was wrong, and the table ended-up with duplicates.
This PR just add a small check to ensure overwrites files actually exists in the current snapshot before committing.

### DIFF
--- a/iceberg-rust/src/table/transaction/operation.rs
+++ b/iceberg-rust/src/table/transaction/operation.rs
@@ -32,7 +32,7 @@ use tokio::task::JoinHandle;
 use tracing::{debug, instrument};
 
 use crate::table::manifest::{FilteredManifestStats, ManifestWriter};
-use crate::table::manifest_list::ManifestListWriter;
+use crate::table::manifest_list::{ManifestListReader, ManifestListWriter};
 use crate::table::transaction::append::append_summary;
 use crate::{
     catalog::commit::{TableRequirement, TableUpdate},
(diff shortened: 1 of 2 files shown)


arnaudbriche:
Hi @JanKaul , is there an issue with this PR ? Can I help somehow ?

JanKaul:
Sorry for the delay. The PR https://github.com/JanKaul/iceberg-rust/pull/292 surfaced a bug in the overwrite operation. And I wanted to thoroughly go over the overwrite operation again but couldn't really find the time to do so.

This looks great. Thanks for the PR.

arnaudbriche:
@JanKaul 

Not problem (:

I've spent some time thinking about the right way to ensure `overwrite` works correctly.
My conclusion is that the API of the operation should change to take only a list of `data_files` to overwrite rather than a mapping of `manifest` -> `data_files`.

Here is why:
    - `overwrite` implementation must ensure that the files to overwrite actually exist in the current snapshot before commiting
    - `overwrite` is mainly used in the context of partial table compaction (unless I'm missing some use-cases)
    - Compaction is usually a fairly long-running operation (a few minutes to a few hours)
    - There is a non-negligible probability that the manifests of the original snapshot will be overwritten while compaction is running
    - Enforcing integrity by checking against the presence of the original manifests is wrong because it can lead to false positives where a commit is rejected while the data files to overwrites are still part of the table, because of manifest rewrites

To prevent false positives, I do a remapping of datafiles to manifest after compaction has finished.
It works, but I adds overhead, and it could be mutualized with the overwrite implementation

PR #300, opened 2026-01-13
Add snapshot-log pruning to ExpireSnapshots operation

arnaudbriche:
The `snapshot-log` metadata's field is never pruned and so grows indefinitely.
I just added a filtering step to `RemoveSnapshots` update  code to ensure `snapshot-log` contains only valid snapshots.


JanKaul:
That's great thanks a lot

PR #302, opened 2026-01-14
Implement OAut2 client-credentials flow.

arnaudbriche:
This PR started when I tried to use the library to interact with an auth-enabled Lakekeeper instance.
For this to work, the catalog client needs to get (and refresh) an access token from an Oauth2 server (not the catalog server).

As I did not want to touch the generated code, I decided to add an abstraction that would allow to implement rewrites to the Configuration object passed to the generated code.
The client credentials flow has been implemented as a Configuration rewriter.

I also add a flag to disabled getting storage_credentials from the catalog server as it doers not work in some non-AWS environments.


JanKaul:
Looks great. Thanks a lot

PR #305, opened 2026-02-08
Upgrade datafusion from 51 to 52 and adapt API changes

cedricziel:
Bumps DataFusion and all sub-crates from 51 to 52.

### API adaptations

- **`ParquetSource`**: `default()` removed; now requires a schema via `new(table_schema)`
- **`FileScanConfigBuilder::new`**: Schema arg removed (now lives in `FileSource`); signature changed from `(url, schema, source)` to `(url, source)`
- **`with_table_partition_cols`**: Removed from `FileScanConfigBuilder`; partition columns are now passed through `TableSchema::new(file_schema, partition_cols)` into `ParquetSource`
- **`with_projection_indices`**: Now returns `Result` instead of `Self`
- **`ColumnStatistics`**: New required `byte_size` field

### Before/After

```rust
// Before (DF 51)
let source = Arc::new(ParquetSource::default());
FileScanConfigBuilder::new(url, file_schema, source)
    .with_table_partition_cols(partition_cols)
    .with_projection_indices(Some(proj))
    .build();

// After (DF 52)
let table_schema = TableSchema::new(file_schema, partition_col_refs);
let source = Arc::new(ParquetSource::new(table_schema));
FileScanConfigBuilder::new(url, source)
    .with_projection_indices(Some(proj))?
    .build();
```

### DIFF
--- a/datafusion_iceberg/tests/equality_delete.rs
+++ b/datafusion_iceberg/tests/equality_delete.rs
@@ -1,8 +1,4 @@
-use datafusion::{
-    arrow::{error::ArrowError, record_batch::RecordBatch},
-    assert_batches_eq,
-    prelude::SessionContext,
-};
+use datafusion::{arrow::error::ArrowError, assert_batches_eq, prelude::SessionContext};
 use datafusion_iceberg::catalog::catalog::IcebergCatalog;
 use duckdb::Connection;
 use futures::stream;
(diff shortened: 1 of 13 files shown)


JanKaul:
Thanks for the PR.

Not sure why linking with duckdb fails. Maybe try updating duckdb version.

Other than that, you neet to run cargo fmt again.

cedricziel:
So duckdb uses too much disk space when compiled from scratch. - I'll try switching from a bundled version to a direkt installation of duckdb in the runner.

cedricziel:
@JanKaul updated. - It now tries to use an installed version of duckdb which we can download in ci. cargo.config references it as well for the linker. I think it's probably the right tradeoff between efficiency, build time and safety to use a pre-built version

JanKaul:
That's awesome thanks for looking into this

datafusion_iceberg/tests/equality_delete.rs
```diff
@@ -25,25 +21,6 @@ use object_store::local::LocalFileSystem;
...
-    let mut buffer = Vec::new();
-    let mut writer = DuckDBWriter::try_new(&mut buffer, &duckdb_batch.schema()).unwrap();
-    writer.write(&duckdb_batch).unwrap();
-    writer.finish().unwrap();
-    drop(writer);
-
-    let mut reader = DataFusionReader::try_new(std::io::Cursor::new(buffer), None).unwrap();
-    reader.next().unwrap().unwrap()
-}
-
```
JanKaul, line 46: I introduced the conversion because otherwise we are forced to use an Arrow version that is supported by DuckDB. I had the issue that DuckDB didn't support the newest Arrow version and I wasn't able to update Arrow/Datafusion without tests failing. The IPC conversion makes it possible to use different arrow versions.

Even when we use the DuckDB repo for the dependency, this issue might arise in the future. I think keeping this conversion allows us to be completely independent.

And if I'm not mistaked it also should affect preinstalling DuckDB, right?

cedricziel: the reason this was removed was because the two are now aligned and the conversion was not necessary anymore.

JanKaul: Yes, but as soon as Datafusion updates the Arrow version and DuckDB does not, we have a conflict again. I would just like to not introduce this dependency. And since it's only for the tests it doesn't affect performance or anything.

cedricziel: brought the test back

JanKaul:
Hey @cedricziel, thanks for putting your effort into this PR. However, I've decided to merge #307 as its a smaller scoped PR and doesn't change the DuckDB bundling. And it turns out the tests also succeed even without the bundling changes and we can keep things simpler for now.
Again thanks a lot for your contribution and sorry for the inconvenience.

I hope this still provides you the necessary functionality.

cedricziel:
cool with me - I'm just here for the upgraded dependencies ;)

PR #308, opened 2026-02-27
Fix decimal value conversion in statistics following DataFusion v52 upgrade.

lizardoluis:
## Problem

After upgrading to DataFusion v52, decimal values in query planning statistics 
were not being converted correctly. The conversion was missing:
1. Proper use of mantissa for decimal scalar values
2. Field type information needed for extracting precision and scale

This led to incorrect decimal representations in statistics and affected query 
planning accuracy.


lizardoluis:
@JanKaul this is some bits left from the upgrade of Datafusion.

JanKaul:
Thanks for the fix.

PR #321, opened 2026-03-20
fix: ensure we only push down filters to delete files that are compatible with their schema

gruuya:
Otherwise we'll get errors for any query where the predicate doesn't involve an equality delete column. 

For instance the new additions to the `test_equality_delete` test would lead to a panic
```
Failed to execute query: Shared(ArrowError(SchemaError("Unable to get field named \"amount\". Valid fields: [\"id\", \"customer_id\", \"product_id\", \"date\"]"), Some("")))
```


JanKaul:
Great work, thanks a lot!

PR #329, opened 2026-03-27
Databricks Unity Catalog support and bug fixes

lizardoluis:
This PR adds support for Databricks Unity Catalog and fixes several issues in the REST catalog implementation.

### 1. Support multi-segment prefixes in REST API calls

Enable REST catalogs to use hierarchical prefix patterns like `catalogs/warehouse_name` by encoding each path segment individually instead of encoding the entire prefix string.

**Before:** `catalogs/warehouse` → `catalogs%2Fwarehouse` (404 errors)  
**After:** `catalogs/warehouse` → `/catalogs/warehouse` (proper path segments)

This is required by Databricks Unity Catalog and other REST catalog implementations that return custom prefixes in their `/v1/config` response.

Note: This is more like a workaround. It would be better instead to check the response returned by the catalog and extract the prefix from there. PyIceberg does like this. 

### 2. Handle missing views endpoint and empty storage credentials

- **Views endpoint:** Databricks unity catalog don't implement the views endpoint yet. Now treats 404 responses as an empty view list instead of failing.
- **Empty credentials:** Added check for empty `storage_credentials` vector to prevent index out of bounds panic when catalogs return an empty array.

### 3. Fix missing path separator in Parquet file paths

Fixed path construction in `generate_file_path()` that was missing a separator between the data location and filename UUID.

**Before:** `tables/{table-uuid}{file-uuid}.parquet`  
**After:** `tables/{table-uuid}/{file-uuid}.parquet`

Adds logic to insert `/` when neither the base path ends with `/` nor the partition path starts with `/`.

### DIFF
--- a/catalogs/iceberg-rest-catalog/src/apis/fetch.rs
+++ b/catalogs/iceberg-rest-catalog/src/apis/fetch.rs
@@ -18,14 +18,7 @@ where
     T: for<'a> serde::Deserialize<'a>,
     E: for<'a> serde::Deserialize<'a>,
 {
-    let uri_base = match prefix {
-        Some(prefix) => format!(
-            "{}/v1/{prefix}/",
-            configuration.base_path,
-            prefix = crate::apis::urlencode(prefix)
-        ),
-        None => format!("{}/v1/", configuration.base_path,),
-    };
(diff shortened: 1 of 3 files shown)


JanKaul:
This is great, thanks a lot!

PR #334, opened 2026-04-21
fix: make value lookup skip nested fields

gruuya:
We noticed that selects were returning zero rows for Iceberg tables with nested columns created via pyspark.

The problem is two-fold:
- (py)spark will dump nested column lower/upper bounds to the manifest files, but when creating the `Value`s map from the avro file this crate uses the top-level schema, so it errors out on nested fields
- datafiles function will silently swallow any parsing issues in the manifest, and consequently the query goes through without warning, but it effectively has no manifests involved in the table state (hence 0 rows in all selects)

This PR fixes both:
- bubble up any errors encountered when parsing the manifests
- skip upper/lower bound values for fields that are not in the top-level schema

### DIFF
--- a/iceberg-rust-spec/src/error.rs
+++ b/iceberg-rust-spec/src/error.rs
@@ -13,9 +13,6 @@ pub enum Error {
     /// Type error
     #[error("Value {0} doesn't have the {1} type.")]
     Type(String, String),
-    /// Schema error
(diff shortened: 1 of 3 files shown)


JanKaul:
Thanks a lot for the fix! Looks good to me

PR #366, opened 2026-06-26
fix: make pruning statistics TZ-aware

gruuya:
Otherwise DataFusion pruning can fail on comparisons with `Only intervals with the same data type are comparable, lhs:Timestamp(µs), rhs:Timestamp(µs, "UTC")`

In addition expand scalar conversion to accomodate timestamptz as well.


JanKaul:
Thanks for the fix. LGTM

PR #368, opened 2026-07-02
fix: decimal and uuid value parsing from parquet types

gruuya:
Currently we erroneously pars stats and partition values for UUID and Decimal types in some cases (e.g. when the underlying Parquet types are `parquet::basic::Type::BYTE_ARRAY` or `parquet::basic::Type::INT32`/`parquet::basic::Type::INT64` respectively).

Consequently partition pruning can be wrong, leading to leaving out some partitions that should otherwise be considered. Another failure mode is when DataFusion relies on the stats to compute some aggregation, and the value is garbage.

As a fix add a hint for the physical types representing the values which is used in `Value::try_from_bytes` to invoke different parsing paths (e.g. from a string or reverse LE to BE)


JanKaul:
I've suggested a small refactor on top of this PR: instead of changing the existing `try_from_bytes` signature (which forces all callers to pass `None`), rename the 3-argument form to `try_from_bytes_with_hint` and restore a clean 2-argument `try_from_bytes` that delegates with `None`. That way only `parquet.rs` and the hint-specific tests need to call the longer form.

See: https://github.com/splitgraph/iceberg-rust/pull/22

gruuya:
> I've suggested a small refactor on top of this PR: instead of changing the existing `try_from_bytes` signature (which forces all callers to pass `None`), rename the 3-argument form to `try_from_bytes_with_hint` and restore a clean 2-argument `try_from_bytes` that delegates with `None`. That way only `parquet.rs` and the hint-specific tests need to call the longer form.
> 
> See: [splitgraph#22](https://github.com/splitgraph/iceberg-rust/pull/22)

Thanks, that's cleaner!

JanKaul:
Thanks a lot for the PR!

PR #369, opened 2026-07-02
fix: accept legacy Athena field names in manifest list parsing

### LINKED ISSUE #367: Robust Avro schema parsing
We ran into an issue where AWS Athena still uses legacy names for some fields in the manifest list entry, namely:

1. `added_data_files_count` instead of `added_files_count` (field id 504)
2. `existing_data_files_count` instead of `existing_files_count` (field id 505)
3. `deleted_data_files_count` instead of `deleted_files_count` (field id 506)

This seems to stem from an early misalignment between the spec and the reference Java implementation https://github.com/apache/iceberg/issues/8684.

So in effect Athena doesn't comply with the spec, and consequently trying to read tables written by it leads to an error: `Missing field in record: \"added_files_count\"`.

While in theory this should be fixed on the Athena side, in practice it's probably just easier to make the fix in this crate.

For instance either of:
1. provide aliases for some fields in the `manifest_list_schema_v*`; the problem here is that avro-rs doesn't seem to support parsing aliases right now.
2. mark these fields as optional; this is also meh, since it further departs from the spec
3. don't supply the Avro reader with a schema at all, make it parse the embedded one and then validate the required field ids are present; likely has some drawbacks too

JanKaul:
## Summary

- AWS Athena (and possibly other engines) writes `added_data_files_count`, `existing_data_files_count`, and `deleted_data_files_count` in manifest list files instead of the spec-correct names — a known early misalignment tracked in [apache/iceberg#8684](https://github.com/apache/iceberg/issues/8684)
- Reading such files failed with `Missing field in record: "added_files_count"` because `apache_avro` performs Avro-level field resolution by name when a reader schema is supplied, and `avro-rs` has no alias support
- Fix: read without a reader schema (using the embedded writer schema) so the raw field names reach the serde layer, then use `#[serde(alias)]` to accept both the legacy and spec-correct names on all three affected fields across V1/V2/V3 serde structs
- A `TODO` comment marks the reader-schema removal as temporary — to be reverted once all major query engines write the spec-correct names

Fixes #367

## Test plan

- [ ] Existing `test_manifest_list_v1/v2/v3` unit tests pass
- [ ] Manually verify reading a manifest list written by Athena resolves without error

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/iceberg-rust-spec/src/spec/manifest_list.rs
+++ b/iceberg-rust-spec/src/spec/manifest_list.rs
@@ -134,10 +134,13 @@ mod _serde {
         /// ID of the snapshot where the manifest file was added
         pub added_snapshot_id: i64,
         /// Number of entries in the manifest that have status ADDED (1)
+        #[serde(alias = "added_data_files_count")]
         pub added_files_count: i32,
         /// Number of entries in the manifest that have status EXISTING (0)
+        #[serde(alias = "existing_data_files_count")]
         pub existing_files_count: i32,
         /// Number of entries in the manifest that have status DELETED (2)
+        #[serde(alias = "deleted_data_files_count")]
         pub deleted_files_count: i32,
         /// Number of rows in all of files in the manifest that have status ADDED
         pub added_rows_count: i64,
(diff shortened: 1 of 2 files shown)


gruuya (approved):
I think this will work, we'd love to try it out.

PR #370, opened 2026-07-06
fix: add handling for Date32 and Decimal stat parsing

gruuya:
Extend `any_iter_to_array` so that we refine pruning on 2 new data types.


JanKaul:
Thanks a lot for the fix. LGTM

PR #371, opened 2026-07-07
fix: expose concurency control for manifest read/write operations

gruuya:
Previously for queries involving many tables/manifests the `datafiles` call would fan out too aggressively, potentially leading to CP throttling etc.

With the new knob the users can control that now.

### DIFF
--- a/iceberg-rust/src/table/mod.rs
+++ b/iceberg-rust/src/table/mod.rs
@@ -12,7 +12,6 @@
 
 use std::{io::Cursor, sync::Arc};
 
-use futures::future::try_join_all;
 use itertools::Itertools;
 use manifest::ManifestReader;
 use manifest_list::read_snapshot;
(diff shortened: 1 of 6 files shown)


iceberg-rust/src/table/mod.rs
```diff
@@ -328,55 +327,51 @@ async fn datafiles(
...
-            };
-
-            let sequence_number = if let Some(sequence_number) = x.sequence_number() {
-                *sequence_number
-            } else {
-                *x.sequence_number_mut() = Some(sequence_number);
-                sequence_number
-            };
+    // Bounded concurrency caps in-flight object-store requests and peak decode memory.
+    let entries: Vec<(ManifestPath, ManifestEntry)> = stream::iter(futures)
```
gruuya: Btw i also have a commit that avoids this intermediate materialization into a vector by changing the return type (and all callers) to be a Stream. Let me know if this is alright to include or not.

JanKaul: You mean that the "datafiles" function return a stream? Yeah I think that would be a great idea.

JanKaul (commented):
Looks good to me.

I left one comment regarding returning a stream. I think it would be a good idea.

iceberg-rust/src/table/mod.rs
gruuya: Yes, exactly; rebased and pushed another commit.

JanKaul:
Thank you

Issue #268, opened 2025-10-28
Add support for REST cache semantics in catalog


gruuya:
Iceberg has recently added [ETag-based](https://github.com/apache/iceberg/pull/11946) caching support to the REST API, which can be useful to avoid always pulling table metadata if nothing has changed.

We'd like to use this eventually with the (REST) catalog client, so I'm wondering what is the best way to integrate this
1. at the very minimum this would involve changing `catalog_api_api::load_table` to support adding additional headers, so that [`If-None_Match`](https://github.com/apache/iceberg/blob/6a0d4a0f37a22a885ef43b37353bee258ca93068/open-api/rest-catalog-open-api.yaml#L942) can be provided, along with
2. extending the results to include the `etag`s in [`LoadTableResponse`](https://github.com/apache/iceberg/blob/6a0d4a0f37a22a885ef43b37353bee258ca93068/open-api/rest-catalog-open-api.yaml#L4738-L4740) and [`CreateTableResponse`](https://github.com/apache/iceberg/blob/6a0d4a0f37a22a885ef43b37353bee258ca93068/open-api/rest-catalog-open-api.yaml#L4707-L4709), and potentially
3. (optional) making the catalog freshness-aware by
    a) making RestCatalog internally freshness-aware, by maintaining its own metadata cache (ideally something that can be shared between different instances) and retaining `Catalog` trait as is, or
    b) extending the `Catalog::load_tabular` method so that the ETag can be passed at this level, and thus delegating caching to the caller

In principle 1 and 2 are required for someone to build a custom freshness-aware REST catalog, but 3 a or b would be nice to abstract over this and have more flexibility.

PR #372, opened 2026-07-13
feat: introduce rest catalog cache primitives

gruuya:
Progresses #268.

This PR adds new primitives for handling the HTTP cache semantics when loading a table:
- new `Conditional` enum in the rest crate
- generalizing fetch, so that it can accommodate both `LoadTableResult` and `Conditional<LoadTableResult>` via 
- a new `FromResponse` trait
- (drive-by) `fetch` and `fetch_empty` are consolidated to remove code duplication.

Finally, this is also exposed at the `catalog_api_api::load_table` level, which now accepts headers, and can return `Conditional::NotModified` when a `If-None-Match` headers is passed, end the ETag is still matching.

### DIFF
--- a/catalogs/iceberg-rest-catalog/src/apis/catalog_api_api.rs
+++ b/catalogs/iceberg-rest-catalog/src/apis/catalog_api_api.rs
@@ -444,7 +444,7 @@ pub async fn commit_transaction(
     let uri_str = "transactions/commit".to_string();
     let method = reqwest::Method::POST;
 
-    fetch::fetch_empty(
+    fetch::fetch(
         configuration,
         method,
         prefix,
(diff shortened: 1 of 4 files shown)


JanKaul:
Thanks a lot for the PR. IT would be great to get this in.

I've made a PR https://github.com/splitgraph/iceberg-rust/pull/23 where a refactored the `FromResponse` trait into a inner and wrapper function. I would like to avoid using a trait for this.

JanKaul:
Sorry for the linting issues I pushed to the other PR.

JanKaul:
Otherwise looks good to me.

gruuya:
> Sorry for the linting issues I pushed to the other PR.

no worries, lint fixes pushed.

PR #378, opened 2026-07-28
fix: resolve current_schema via current_schema_id, not the snapshot pin

### LINKED ISSUE #376: Schema evolution: Append/Replace snapshots re-pin the pre-evolution schema_id, so SetCurrentSchema never takes effect
Related to #303. We (SignalDB) drive schema evolution through the low-level commit path, and found that adding a column can never become effective:

1. `Catalog::update_table` with `TableUpdate::AddSchema { schema: v1, .. }` + `TableUpdate::SetCurrentSchema { schema_id: 1 }` commits fine — the resulting metadata has `schemas {0, 1}` and `current_schema_id = 1`.
2. But `TableMetadata::current_schema(branch)` prefers the **current snapshot's** pinned `schema_id` over `current_schema_id` (`iceberg-rust-spec/src/spec/table_metadata.rs` ~L162):

```rust
let schema_id = self
    .current_snapshot(branch)?
    .and_then(|x| *x.schema_id())
    .unwrap_or(self.current_schema_id);
```

3. And the `Append`/`Replace` transaction operations stamp their **new** snapshot's `schema_id` from `current_schema(branch)` (`table/transaction/operation.rs` ~L265 append, ~L616 replace).

So on any table that already has a snapshot: the old snapshot pins schema 0 → `current_schema()` returns schema 0 → the next append/replace stamps its snapshot with schema 0 → and so on. The `SetCurrentSchema` flip is unreachable; writers (`write_parquet_partitioned` derives its Arrow schema from `current_schema`) keep producing the old shape forever.

**Proposed fix:** stamp new snapshots with `metadata.current_schema_id` instead of `current_schema(branch)`. Per the Iceberg spec, a snapshot's `schema-id` records "the table schema at the time the snapshot was created" — i.e. the *current* schema id at write time — while the table's authoritative current schema is `current-schema-id`. Arguably `current_schema()` should also resolve via `current_schema_id` (snapshot schema ids are for time-travel reads of that snapshot).

Repro: an (ignored) integration test demonstrating the sequence lives in our tree: https://github.com/cedricziel/signaldb/blob/main/tests-integration/tests/writer/iceberg_integration.rs (`test_schema_evolution_add_label_column`); full analysis: https://github.com/cedricziel/signaldb/issues/734#issuecomment-5109570517

Happy to send a PR if you agree with the direction. (Tested at rev 170ef972.)

cedricziel:
Fixes #376.

`TableMetadata::current_schema` preferred the current snapshot's pinned `schema_id` over `current_schema_id`. That made `SetCurrentSchema` unreachable on any table that already has a snapshot: the old snapshot pins the old schema id, and the `Append`/`Replace` operations stamp their new snapshot's `schema_id` from `current_schema()`, re-pinning the old schema on every commit — so a schema evolution could never take effect and writers (which derive their Arrow schema from `current_schema`) kept producing the old shape.

Per the Iceberg spec, the table's authoritative schema is `current-schema-id`; a snapshot's `schema-id` records the schema that was current *when that snapshot was written*, and stays reachable via `TableMetadata::schema(snapshot_id)` for snapshot-scoped reads (time travel is unaffected).

With this change the four snapshot-stamping sites in `operation.rs` (which all call `current_schema()`) automatically stamp new snapshots with the current schema id, so the AddSchema → SetCurrentSchema → write sequence works end to end.

Includes a regression test: metadata with `schemas {0, 1}`, `current-schema-id: 1`, and a snapshot pinned to schema 0 — `current_schema(None)` must return schema 1 while `schema(snapshot_id)` still returns schema 0.

Context: SignalDB drives schema evolution from its compaction path (cedricziel/signaldb#734) and hit this as a hard blocker.

### DIFF
--- a/iceberg-rust-spec/src/spec/table_metadata.rs
+++ b/iceberg-rust-spec/src/spec/table_metadata.rs
@@ -162,14 +162,10 @@ impl TableMetadata {
     /// stamped its snapshot from this method, re-pinning the old schema
     /// forever.
     ///
-    /// # Arguments
-    /// * `_branch` - Retained for API compatibility; the current schema is a
-    ///   table-level property and does not vary by branch
-    ///
     /// # Returns
     /// * `Result<&Schema, Error>` - The current schema, or an error if the schema cannot be found
     #[inline]
-    pub fn current_schema(&self, _branch: Option<&str>) -> Result<&Schema, Error> {
+    pub fn current_schema(&self) -> Result<&Schema, Error> {
         self.schemas
             .get(&self.current_schema_id)
             .ok_or_else(|| Error::InvalidFormat("schema".to_string()))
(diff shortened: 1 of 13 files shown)


iceberg-rust-spec/src/spec/table_metadata.rs
```diff
@@ -151,21 +151,27 @@ pub struct TableMetadata {
...
     ///
     /// # Returns
     /// * `Result<&Schema, Error>` - The current schema, or an error if the schema cannot be found
     #[inline]
-    pub fn current_schema(&self, branch: Option<&str>) -> Result<&Schema, Error> {
-        let schema_id = self
-            .current_snapshot(branch)?
-            .and_then(|x| *x.schema_id())
-            .unwrap_or(self.current_schema_id);
+    pub fn current_schema(&self, _branch: Option<&str>) -> Result<&Schema, Error> {
```
JanKaul: I think it would make sense to drop the `branch` parameter entirely.

JanKaul (commented):
Thanks a lot the PR. The changes look good. I left a minor comment.

PR #379, opened 2026-07-28
feat: honor per-column bloom-filter table properties in the Parquet writer

### LINKED ISSUE #377: write_parquet_partitioned hardcodes WriterProperties — no way to enable per-column Parquet bloom filters
`store_parquet_partitioned` builds its `WriterProperties` inline (`iceberg-rust/src/arrow/write.rs` ~L522):

```rust
let mut props_builder =
    WriterProperties::builder().set_compression(Compression::ZSTD(ZstdLevel::try_new(1)?));
```

Only `write.metadata.metrics.distinct-counts.enabled` is consulted from table properties. There is no way for a caller to enable Parquet split-block bloom filters (or otherwise tune the writer) — `set_bloom_filter_enabled` / per-column `set_column_bloom_filter_enabled` are unreachable.

**Proposal:** honor the standard Iceberg table properties in the builder, e.g.

- `write.parquet.bloom-filter-enabled.column.<name>` = `true|false` → `set_column_bloom_filter_enabled(ColumnPath, bool)`
- (optionally) `write.parquet.bloom-filter-max-bytes`, `write.parquet.compression-codec` / `-level`

Callers can already reach table properties via the existing `update_properties` transaction op, so this closes the loop without any API change — properties set on the table would just take effect on the next write.

Context: SignalDB wants per-column blooms on materialized label columns for point-lookup pruning (https://github.com/cedricziel/signaldb/issues/731). Happy to send a PR. (Rev 170ef972.)

cedricziel:
Fixes #377.

`store_parquet_partitioned` built its `WriterProperties` inline with only compression set, so callers had no way to enable Parquet split-block bloom filters. This honors the standard Iceberg property `write.parquet.bloom-filter-enabled.column.<name>` = `true|false` per column (new `WRITE_PARQUET_BLOOM_FILTER_ENABLED_COLUMN_PREFIX` const in `iceberg-rust-spec`).

Table properties are already settable through the existing `update_properties` transaction op, so no API change is needed — set the property on a table and the next write picks it up. Unrelated properties are ignored; anything but `"true"` disables the filter for that column.

Includes a unit test (`bloom_filter_properties_apply_per_column`).

Context: SignalDB wants blooms on its materialized label columns for point-lookup pruning (cedricziel/signaldb#731).

### DIFF
--- a/iceberg-rust/src/arrow/write.rs
+++ b/iceberg-rust/src/arrow/write.rs
@@ -551,9 +551,13 @@ fn apply_bloom_filter_properties(
 ) -> parquet::file::properties::WriterPropertiesBuilder {
     for (key, value) in table_properties {
         if let Some(column) = key.strip_prefix(WRITE_PARQUET_BLOOM_FILTER_ENABLED_COLUMN_PREFIX) {
-            let enabled = value == "true";
-            props_builder =
-                props_builder.set_column_bloom_filter_enabled(ColumnPath::from(column), enabled);
+            let enabled = value.eq_ignore_ascii_case("true");
+            // Parquet addresses nested columns by path parts; a dotted name
+            // passed as one string would be treated as a single segment and
+            // silently never match.
+            let path_parts: Vec<String> = column.split('.').map(String::from).collect();
+            props_builder = props_builder
+                .set_column_bloom_filter_enabled(ColumnPath::from(path_parts), enabled);
         }
     }
     props_builder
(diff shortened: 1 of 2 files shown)


iceberg-rust/src/arrow/write.rs
```diff
@@ -538,6 +540,25 @@ async fn create_arrow_writer(
...
+/// per column. Unrelated properties are ignored.
+fn apply_bloom_filter_properties(
+    mut props_builder: parquet::file::properties::WriterPropertiesBuilder,
+    table_properties: &HashMap<String, String>,
+) -> parquet::file::properties::WriterPropertiesBuilder {
+    for (key, value) in table_properties {
+        if let Some(column) = key.strip_prefix(WRITE_PARQUET_BLOOM_FILTER_ENABLED_COLUMN_PREFIX) {
+            let enabled = value == "true";
+            props_builder =
+                props_builder.set_column_bloom_filter_enabled(ColumnPath::from(column), enabled);
```
JanKaul: `ColumnPath::from(column)` is wrong for nested fields

`ColumnPath::from(&str)` wraps the entire string as a single path segment, so "my_struct.label_env" becomes ColumnPath { parts: ["my_struct.label_env"] } instead of ColumnPath { parts: ["my_struct", "label_env"] }. The bloom filter silently has no effect on any nested column. The test doesn't catch this because it only covers flat schemas.

Fix:
```
let path_parts: Vec<String> = column.split('.').map(String::from).collect();
props_builder = props_builder.set_column_bloom_filter_enabled(ColumnPath::from(path_parts), enabled);
```

iceberg-rust/src/arrow/write.rs
```diff
@@ -538,6 +540,25 @@ async fn create_arrow_writer(
...
+/// Honors the standard Iceberg property
+/// `write.parquet.bloom-filter-enabled.column.<name>` = `true`/`false`
+/// per column. Unrelated properties are ignored.
+fn apply_bloom_filter_properties(
+    mut props_builder: parquet::file::properties::WriterPropertiesBuilder,
+    table_properties: &HashMap<String, String>,
+) -> parquet::file::properties::WriterPropertiesBuilder {
+    for (key, value) in table_properties {
+        if let Some(column) = key.strip_prefix(WRITE_PARQUET_BLOOM_FILTER_ENABLED_COLUMN_PREFIX) {
+            let enabled = value == "true";
```
JanKaul: Could we use a case-insensitive chech here?

JanKaul (commented):
Thanks for the PR. Generally looks good to me, I left 2 smaller comments

PR #380, opened 2026-07-31
fix: use full table schema when producing the prune map

gruuya:
Instead of the partition schema, which contains only a subset of columns for partitioned tables in general, or no columns at all for unpartitioned tables, compute the prune map with the full table arrow schema instead.

Note that this didn't impact correctness, since row filtering was still applied correctly and moreover the files were pruned dynamically after inspecting their stats during execution. it's just that unpartitioned tables got zero plan-time file pruning and partitioned tables got none on non-partition columns. 

Instead, now we just skip processing those files at plan time altogether, which can be especially useful for tables with many files (previously we payed a footer fetch during execution for each one just to discard it based on the stats).


JanKaul:
Hmm, I was trying to think of a reason why I only applied the pruning to the partition columns. Doesn't make sense. Great catch, thanks a lot.

PR #382, opened 2026-08-03
feat(sql-catalog): reclaim superseded metadata files on commit

cedricziel:
## Problem

`SqlCatalog::update_table` writes a new `metadata.json` on every commit but never maintains the metadata log or deletes the superseded file. Over a long-lived, frequently-committed table this accumulates one orphaned `metadata.json` per commit in the object store, unbounded. The standard Iceberg table properties that bound this — `write.metadata.previous-versions-max` and `write.metadata.delete-after-commit.enabled` — were not honored.

## Change

In `update_table`:

- **Maintain `metadata_log`**: append the superseded metadata file (with the previous metadata's `last-updated-ms`) on each commit — previously `metadata_log` was left empty.
- **Honor `write.metadata.previous-versions-max`** (default 100): trim the log to the retained count.
- **Honor `write.metadata.delete-after-commit.enabled`** (default false): when enabled, delete the metadata files trimmed from the log.

Deletion happens only after the compare-and-swap `UPDATE` reports `rows_affected() > 0`, so if a concurrent writer wins the CAS we never delete files the winner may still reference. Deletion is best-effort — a failed delete only leaves an orphan, never corrupts state. Default behavior is unchanged (the property is off by default).

## Test

Adds `delete_after_commit_prunes_old_metadata_files`: an in-memory sqlite catalog + `InMemory` object store (no containers) that enables the property with `previous-versions-max = 1`, performs several commits, and asserts the `metadata.json` files are pruned to the current + one previous rather than one-per-version.

## Notes

Only the SQL catalog's table path is covered here; views and the REST/file catalogs are out of scope for this PR. Happy to extend if you'd like it broader.

### DIFF
--- a/catalogs/iceberg-sql-catalog/src/lib.rs
+++ b/catalogs/iceberg-sql-catalog/src/lib.rs
@@ -21,7 +21,7 @@ use iceberg_rust::{
     object_store::{store::IcebergStore, Bucket, ObjectStoreBuilder},
     spec::{
         materialized_view_metadata::MaterializedViewMetadata,
-        table_metadata::{new_metadata_location, TableMetadata},
+        table_metadata::{new_metadata_location, MetadataLog, TableMetadata},
         tabular::TabularMetadata,
         util::strip_prefix,
         view_metadata::ViewMetadata,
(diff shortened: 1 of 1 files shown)


cedricziel:

JanKaul:
Sorry for the delay. I had to think about this one. Wasn't sure whether some people would like to still see the files after.

Anyway, thanks for the PR.

PR #383, opened 2026-08-06
fix(manifest): zero, not absent, added counts on a filtered manifest

cedricziel:
`ManifestWriter::from_existing_with_filter` rewrites every surviving entry with status `Existing` and then clears the manifest's "added" counters by setting them to `None`. `from_existing` does the same.

`added_files_count` and `added_rows_count` are REQUIRED fields in the v2/v3 manifest-list schema, and `ManifestListEntryV2::from` unwraps them. A manifest that adds nothing must therefore say so with zero rather than an absent value: leaving them `None` panics as soon as the entry is serialized back into the manifest list.

`from_existing` escapes this because the append path calls `add_file` afterwards, which repopulates the counters. `from_existing_with_filter` does not — the manifest is finished immediately — so every `overwrite` whose filtered manifest still holds a surviving entry panics in `append_and_filter`.

`test_table_transaction_overwrite` misses it because it removes every file its filtered manifest holds, leaving the manifest empty; `append_and_filter` then skips serialization via its `added > 0 || existing > 0` guard. The bug needs a filtered manifest that is *not* the one selected to receive the new data files (that one gets `add_file` called on it) and that still holds a survivor — which is exactly the shape of a partition-scoped compaction that merges a subset of a table's files.

The added regression test reproduces that: 32 single-file appends into one partition, enough for the manifest list to split, then an overwrite removing one file from each multi-entry manifest. It panics at `manifest_list.rs:294` without this change and passes with it.

Both call sites are fixed so the invariant — a v2 manifest-list entry never carries an absent required count — holds uniformly.

### DIFF
--- a/iceberg-rust/src/table/manifest.rs
+++ b/iceberg-rust/src/table/manifest.rs
@@ -393,8 +393,15 @@ impl<'schema, 'metadata> ManifestWriter<'schema, 'metadata> {
             manifest.existing_rows_count.unwrap_or(0) + manifest.added_rows_count.unwrap_or(0),
         );
 
-        manifest.added_files_count = None;
-        manifest.added_rows_count = None;
+        // Zero, not absent: `added_files_count` / `added_rows_count` are
+        // REQUIRED fields in the v2/v3 manifest-list schema, and
+        // `ManifestListEntryV2::from` unwraps them. A manifest whose entries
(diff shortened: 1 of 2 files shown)


JanKaul:
Great fix, thanks for the PR!

PR #384, opened 2026-08-06
fix(sql-catalog): report a conflict when the metadata swap matches no rows

cedricziel:
`update_table`/`update_view`/`update_materialized_view` swap the catalog's metadata pointer with a compare-and-swap that only matches while the base location is still current, but the result was discarded. A writer whose base had been superseded by a concurrent commit returned `Ok` after updating zero rows — the new `metadata.json` was written but nothing referenced it, and the caller was told it succeeded.

Reachable whenever two catalog instances share a database: each has its own cache, so the requirement check passes against a stale base and the CAS is the only guard left.

Returns a new `Error::CommitConflict` when the update matches no rows, and drops the stale cache entry so the next load refreshes. The three identical statements move to one helper.

Regression test: two `SqlCatalog` instances over one database; the loser's commit must error and must not overwrite the winner.


JanKaul:
Thanks a lot for the PR!

PR #385, opened 2026-08-06
feat: honor the write.metadata.metrics.* table properties

cedricziel:
Bounds were collected for every column at full length. Since statistics live inline in the manifest entry for every data file, a table with a JSON blob or a log body carried two untruncated copies of its longest and shortest value on every entry, forever — to prune nothing.

Honors the spec's controls, with the four modes `none` / `counts` / `truncate(n)` / `full`:

| Property | Default |
|---|---|
| `write.metadata.metrics.default` | `truncate(16)` |
| `write.metadata.metrics.column.<name>` | — |
| `write.metadata.metrics.max-inferred-column-defaults` | `100` |

Resolution is per-column override → default → inferred default, and the inferred default only reaches the first `max-inferred-column-defaults` top-level columns.

Details worth a look:

- Truncation runs **once**, after bounds are merged across row groups. Per-row-group truncation would repeatedly widen an already-widened upper bound.
- Lower bounds truncate to a prefix (can only lower them). Upper bounds truncate then increment, so they still cover every value; when the prefix cannot be incremented the bound is **dropped** rather than understated — an understated upper bound prunes files that hold matching rows.
- String bounds truncate by character, never splitting one; increments skip the surrogate range.
- `column_sizes` describes layout rather than values, so it stays unconditional (matches the Java implementation).

**Behavior change:** the default becomes `truncate(16)` instead of full bounds. Set `write.metadata.metrics.default = "full"` to keep the old behavior.

Tests: 17 unit tests for parsing, resolution and truncation invariants, plus 3 end-to-end tests through `parquet_to_datafile` (default truncation, `none` dropping a column's metrics, `full` keeping untruncated bounds).

### DIFF
--- a/iceberg-rust-spec/src/spec/table_metadata.rs
+++ b/iceberg-rust-spec/src/spec/table_metadata.rs
@@ -53,6 +53,16 @@ pub const WRITE_METADATA_METRICS_DISTINCT_COUNTS_ENABLED: &str =
 /// `write.parquet.bloom-filter-enabled.column.label_env = "true"`.
 pub const WRITE_PARQUET_BLOOM_FILTER_ENABLED_COLUMN_PREFIX: &str =
     "write.parquet.bloom-filter-enabled.column.";
+/// Default statistics mode for a table's columns: `none`, `counts`,
+/// `truncate(n)` or `full`. Defaults to `truncate(16)`.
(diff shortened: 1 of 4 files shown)


JanKaul:
Yet, anther great PR, thanks!

PR #386, opened 2026-08-06
feat(sql-catalog): let callers configure the connection pool and its sessions

cedricziel:
`SqlCatalog::new` opens its pool through sqlx's `Any` driver, which takes a URL and nothing else. A caller can neither size the pool nor configure the sessions on it.

Both matter under load:

- The pool default lets ten writers contend on a database that may serialize them.
- Per-session settings — SQLite's `busy_timeout`, PostgreSQL's `statement_timeout` — cannot be carried on the URL either; sqlx's SQLite URL parser rejects them as query parameters. Today the only workaround is to open a separate connection out of band and hope the setting happens to be persistent.

Adds `SqlCatalogOptions` plus `new_with_options` on `SqlCatalog` and `SqlCatalogList`:

```rust
let options = SqlCatalogOptions::new()
    .with_pool_options(PoolOptions::new().max_connections(4))
    .with_session_statements(["pragma busy_timeout = 30000"]);
let catalog = SqlCatalog::new_with_options(url, "warehouse", store, options).await?;
```

`new` delegates with the defaults, so existing callers are unaffected — including the single-connection cap that keeps a private in-memory SQLite database from being opened twice. Both constructors had copies of the same connection setup; it moves to one helper. `SqlCatalogList` still connects eagerly and `SqlCatalog` lazily, so the helper returns configured options rather than a pool.

Tests: session statements reach pooled connections, pool options are honored, and the old constructor keeps its previous behavior end to end.

### DIFF
--- a/catalogs/iceberg-sql-catalog/src/lib.rs
+++ b/catalogs/iceberg-sql-catalog/src/lib.rs
@@ -109,14 +109,14 @@ impl SqlCatalog {
     ) -> Result<Self, Error> {
         install_default_drivers();
 
-        let pool_options = options.pool.unwrap_or_else(|| {
-            let pool_options = PoolOptions::new();
-            if url == "sqlite://" {
-                pool_options.max_connections(1)
-            } else {
-                pool_options
-            }
-        });
+        let mut pool_options = options.pool.unwrap_or_default();
+        if url == "sqlite://" {
+            // A private in-memory SQLite database is only reachable through the
+            // connection that created it; a second connection opens a distinct,
+            // empty database. Enforce the single-connection cap even when the
+            // caller supplied their own pool options.
+            pool_options = pool_options.max_connections(1);
+        }
 
         let pool = pool_options_with_setup(
             pool_options,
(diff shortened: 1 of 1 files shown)


cedricziel:
Closing/reopening to check whether this nudges GitHub Actions — the current head has had zero check-runs since the force-push, consistent with the ongoing Actions incident (githubstatus.com).

cedricziel:

catalogs/iceberg-sql-catalog/src/lib.rs
```diff
@@ -48,61 +48,81 @@ pub struct SqlCatalog {
...
-        if url == "sqlite://" {
-            pool_options = pool_options.max_connections(1);
-        }
+        let pool_options = options.pool.unwrap_or_else(|| {
+            let pool_options = PoolOptions::new();
+            if url == "sqlite://" {
+                pool_options.max_connections(1)
+            } else {
+                pool_options
+            }
```
JanKaul: I think it makes sense to always set `max_connections` to 1 for sqlite, also when `pool` is Some and not just in the `unwrap_or_else` case. Can we pull this out here?

JanKaul (commented):
Thanks a lot for the PR. Generally looks good to me, I just have one minor comment regarding the sqlite max_connections.

PR #387, opened 2026-08-06
feat: honor the write.parquet.* writer properties

cedricziel:
The Parquet writer ignored the table properties that describe how to write Parquet.

Compression was hardcoded to zstd level 1 — while `CreateTableBuilder` records `write.parquet.compression-codec = zstd` and `write.parquet.compression-level = 3` on every new table, so the metadata described a file that was never written. Page, page-row and dictionary sizing were not reachable at all. And a bloom filter could only be turned on, never sized: it took the default false-positive rate whatever the column's cardinality, so a high-cardinality column got either wasted reads or a needlessly large footer.

Now honored:

| Property | Maps to |
|---|---|
| `write.parquet.compression-codec` | `set_compression` (uncompressed/snappy/gzip/lz4/lz4_raw/brotli/zstd) |
| `write.parquet.compression-level` | the codec's level, where it takes one |
| `write.parquet.page-size-bytes` | `set_data_page_size_limit` |
| `write.parquet.page-row-limit` | `set_data_page_row_count_limit` |
| `write.parquet.dict-size-bytes` | `set_dictionary_page_size_limit` |
| `write.parquet.bloom-filter-fpp.column.<name>` | `set_column_bloom_filter_fpp` |

An unknown codec, unparsable number, a level a codec rejects, or an fpp outside `(0, 1)` is ignored rather than failing the write — a bad property degrades to the default. A level on a codec that takes none is ignored, matching the Java implementation.

**Deliberately not covered:** `write.parquet.row-group-size-bytes` has no parquet-rs equivalent (`set_max_row_group_size` counts rows, not bytes), so honoring it needs byte accounting in the write loop. `write.parquet.bloom-filter-max-bytes` has no equivalent either — parquet-rs sizes a filter from ndv and fpp.

Tests: 6 covering each property, the ignore-and-default paths, and the levelless-codec case.

### DIFF
--- a/iceberg-rust-spec/src/spec/table_metadata.rs
+++ b/iceberg-rust-spec/src/spec/table_metadata.rs
@@ -68,6 +68,8 @@ pub const WRITE_PARQUET_DICT_SIZE_BYTES: &str = "write.parquet.dict-size-bytes";
 /// Maximum size of a Parquet row group in bytes. Row groups are flushed once
 /// their estimated encoded size crosses this, independent of row count.
 pub const WRITE_PARQUET_ROW_GROUP_SIZE_BYTES: &str = "write.parquet.row-group-size-bytes";
+/// Parquet page format: `v1` (DataPage) or `v2` (DataPageV2). Defaults to `v1`.
+pub const WRITE_PARQUET_PAGE_VERSION: &str = "write.parquet.page-version";
 /// Default statistics mode for a table's columns: `none`, `counts`,
 /// `truncate(n)` or `full`. Defaults to `truncate(16)`.
 pub const WRITE_METADATA_METRICS_DEFAULT: &str = "write.metadata.metrics.default";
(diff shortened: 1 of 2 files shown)


cedricziel:
Trial close/reopen — GitHub Actions incident appears to be recovering; checking whether this now triggers a CI run.

iceberg-rust-spec/src/spec/table_metadata.rs
```diff
@@ -53,6 +53,18 @@ pub const WRITE_METADATA_METRICS_DISTINCT_COUNTS_ENABLED: &str =
 /// `write.parquet.bloom-filter-enabled.column.label_env = "true"`.
 pub const WRITE_PARQUET_BLOOM_FILTER_ENABLED_COLUMN_PREFIX: &str =
     "write.parquet.bloom-filter-enabled.column.";
+/// Per-column Parquet bloom-filter false-positive probability: append the
+/// column name, e.g. `write.parquet.bloom-filter-fpp.column.trace_id = "0.01"`.
+/// A filter is sized from this, trading footer bytes against wasted reads.
+pub const WRITE_PARQUET_BLOOM_FILTER_FPP_COLUMN_PREFIX: &str =
+    "write.parquet.bloom-filter-fpp.column.";
```
JanKaul, line 60: How difficult would it be to include the `write.parquet.bloom-filter-ndv.column.`?

If it's not too difficult it would be great to include it in the PR. Otherwise it would be great to leave a comment that it's currently not covered.

JanKaul (commented):
Thanks a lot, this is great stuff. I have one minor remark so that we don't forget the `bloom-filter-ndv` parameter.

Really appreciate it.

iceberg-rust-spec/src/spec/table_metadata.rs
cedricziel: @JanKaul I added the remaining supported properties. Thanks for the fast reviews - appreciate it a lot!

JanKaul: Awesome, thank you

PR #388, opened 2026-08-06
feat: honor write.target-file-size-bytes and roll on real file size

cedricziel:
Files were rolled on a guess.

The threshold compared the *uncompressed* bytes handed to the writer against `COMPRESSION_FACTOR * MAX_PARQUET_SIZE` — 200 × 512 MB, i.e. **102 GB of input**, on the assumption that Parquet compresses 200:1. Real data compresses nowhere near that, so the effective file size was whatever 102 GB of input happened to produce: far past the 512 MB the constant named, and unpredictable, since it moved with how well the data compressed.

Rolls on the writer's actual on-disk size instead — `bytes_written()` for what is flushed, plus `in_progress_size()` for the row group still buffered — against `write.target-file-size-bytes`, defaulting to the spec's 512 MiB. No compression factor is needed once the size being bounded is measured rather than predicted.

The check stays *before* the batch is written, so every file receives at least one batch and no empty file is emitted. `bytes_written` was doing double duty as the "did anything get written" sentinel; that becomes a row count, which is what it was standing in for.

**Behavior change:** files now actually roll at 512 MB, where before they grew until the input crossed 102 GB. `record_batch_size` fed the old guess and has no remaining callers, so it goes.

Tests: default, explicit, and fall-back-on-unusable for the property.

### DIFF
--- a/iceberg-rust-spec/src/spec/table_metadata.rs
+++ b/iceberg-rust-spec/src/spec/table_metadata.rs
@@ -98,6 +98,8 @@ pub const WRITE_METADATA_METRICS_COLUMN_PREFIX: &str = "write.metadata.metrics.c
 /// it are only measured when named explicitly. Defaults to 100.
 pub const WRITE_METADATA_METRICS_MAX_INFERRED_COLUMN_DEFAULTS: &str =
     "write.metadata.metrics.max-inferred-column-defaults";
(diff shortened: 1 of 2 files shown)


cedricziel:

JanKaul:
Thanks a lot, looks good to me.

PR #389, opened 2026-08-06
feat: write gzipped metadata when the table asks for it

cedricziel:
`write.metadata.compression-codec` was half-implemented. `parse_metadata` has long decompressed a `*.gz.metadata.json`, but nothing ever *wrote* one — so the property did nothing, and the decompression path was unreachable in practice.

`metadata.json` is the largest artifact a busy table rewrites on every commit, and it is highly repetitive JSON, so leaving it uncompressed is a standing cost on every commit and every read.

- `new_metadata_location` appends `.gz.metadata.json` when the table's properties name gzip
- `put_metadata` compresses when the location says so

The file name is the single source of truth for the encoding — already how the reader decides — so the two halves cannot disagree.

Anything other than `gzip`, including the spec's `none` default and an unrecognized codec, writes plain JSON. Guessing at an unknown codec would produce a file named as something it is not, which no reader could open.

Adds `TabularMetadataRef::properties`, since choosing the name needs them. The version-hint regex already accepted the `gz.` variant.

Tests: gzip round-trips through writer and reader (asserting the bytes are really a gzip stream and smaller than the JSON), a plain name stays plain, and codec recognition across casings and non-gzip values.

### DIFF
--- a/catalogs/iceberg-file-catalog/src/lib.rs
+++ b/catalogs/iceberg-file-catalog/src/lib.rs
@@ -396,8 +396,11 @@ impl Catalog for FileCatalog {
             .put_metadata(&temp_metadata_location, metadata.as_ref())
             .await?;
 
-        let metadata_location =
-            new_filesystem_metadata_location(&metadata.location, &previous_metadata_location)?;
+        let metadata_location = new_filesystem_metadata_location(
+            &metadata.location,
+            &previous_metadata_location,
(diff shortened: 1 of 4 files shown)


cedricziel:

JanKaul:
Thanks for PR. That's a great feature. I've made some additions to how the file catalog handles the gzip.

PR #390, opened 2026-08-07
fix(append): don't panic when a datafile split has an empty side

cedricziel:
## Problem

`split_datafiles_once` builds a bounding `Rectangle` for each side of the split lazily, then unwraps both:

```rust
smaller_rect.expect("No files selected for the smaller rectangle"),
larger_rect.expect("No files selected for the smaller rectangle"),
```

Both start as `None` and are only populated inside the entry loop. If every manifest entry falls closer to the same bound — or the input iterator is empty — the opposite side never gets a rectangle and the unwrap panics.

This is reachable in normal operation. A partitioned table whose files have been progressively deleted ends up with survivors clustered near one bound of a rectangle that still spans the original range, so every entry sorts to the same side. In our case a routine retention run turned into a crash loop: the crate is built with `panic = "abort"`, so the panic takes the whole process down rather than failing one task.

Worth noting the caller already anticipates empty sides:

```rust
// Filter out empty splits to avoid creating manifests with no entries
```

It just never gets the chance, because the panic fires one frame lower.

(Minor: the `larger_rect` arm also carries the `smaller` message, so the panic text misidentifies which side was empty.)

## Fix

Fall back to the empty `Rectangle` that this same function already uses for its degenerate equal-bounds split a few lines above:

```rust
return Ok([
    (smaller, Rectangle::new(SmallVec::new(), SmallVec::new())),
    (larger, Rectangle::new(SmallVec::new(), SmallVec::new())),
]);
```

An empty rectangle compares `Ordering::Equal` under `cmp_with_priority` (zip over empty slices), so on recursion it re-enters that degenerate branch instead of splitting by position, and `split_datafiles` drops the empty group at the leaves. No manifest is written for a side with no entries.

This keeps the signature unchanged and introduces no new concept — it reuses the crate's existing "no meaningful bounds" sentinel.

## Tests

Four tests in `split_tests`, written before the fix and confirmed to reproduce the panic at the exact `expect` sites:

- `all_entries_on_one_side_yields_an_empty_opposite_split` — the direct regression
- `one_sided_split_drops_the_empty_group` — `split_datafiles` filters the empty side
- `one_sided_split_recurses_without_panicking` — `n_split = 2`, no entries lost
- `empty_input_produces_no_groups` — empty iterator

`cargo test -p iceberg-rust --lib`: 134 passed, 0 failed. `cargo fmt --check` clean, no new clippy warnings.

### DIFF
--- a/iceberg-rust/src/table/transaction/append.rs
+++ b/iceberg-rust/src/table/transaction/append.rs
@@ -62,15 +62,16 @@ fn split_datafiles_once(
             }
         }
     }
+    // A side can legitimately end up empty: every entry may fall closer to the
+    // same bound, and the input iterator may be empty outright. Such a side has
+    // no bounding rectangle, so fall back to the empty one already used for the
+    // degenerate split above. Feeding it back into `split_datafiles_once` hits
+    // that same degenerate branch, and `split_datafiles` drops empty groups.
+    let empty_rect = || Rectangle::new(SmallVec::new(), SmallVec::new());
+
     Ok([
-        (
-            smaller,
-            smaller_rect.expect("No files selected for the smaller rectangle"),
-        ),
-        (
-            larger,
-            larger_rect.expect("No files selected for the smaller rectangle"),
-        ),
+        (smaller, smaller_rect.unwrap_or_else(empty_rect)),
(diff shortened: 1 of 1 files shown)


JanKaul:
Great catch, thank you.

PR #392, opened 2026-08-21
chore: DataFusion 55 upgrade

gruuya:
Bump datafusion crates 54 -> 55, arrow 58 -> 59, and parquet 58 -> 59 (object_store stays on 0.13).

Follow-on API changes required by the bump:
- TableReference moved out of datafusion_sql; import it from datafusion::common instead.
- ExtensionPlanner::plan_extension and QueryPlanner::create_physical_plan now take `&dyn Session` plus a `&PhysicalPlanningContext`, threaded through accordingly in the Iceberg DDL and materialized-view planners.
- ExecutionPlan gained a required `apply_expressions` method; implemented a no-op version for PhysicalForkNode, which carries no expressions.
- create_physical_expr takes an extra `&PhysicalPlanningContext` argument; callers outside of physical planning pass `PhysicalPlanningContext::default()`.
- PruningPredicate::try_new was replaced by PruningPredicateBuilder.
- CreateExternalTable::location became `locations: Vec<String>`; Iceberg CREATE TABLE requires exactly one location.
- TableSchema::new and TableScan::try_new were replaced by their builder equivalents.
- PartitionedFile gained an `arrow_schema` field (set to None, preserving prior behavior of deriving the schema from file metadata).
- SessionConfig's minimum_parallel_output_files / soft_max_rows_per_output_file are now ConfigNonZeroUsize instead of usize.
- parquet 59 deprecated ParquetObjectReader; replaced its one use in iceberg-rust's arrow reader with a small AsyncFileReader implementation over ObjectStore, since the file size is always known from the manifest entry.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -89,7 +89,7 @@ dependencies = [
  "digest 0.10.7",
  "log",
  "miniz_oxide",
- "num-bigint",
+ "num-bigint 0.4.6",
  "quad-rand",
  "rand 0.9.2",
  "regex-lite",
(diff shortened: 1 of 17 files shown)


JanKaul:
A couple of follow-ups on the upgrade:

**`iceberg-rust-spec` dep: keep `arrow-schema` instead of full `arrow`**

The spec crate only references `arrow::datatypes::{DataType, Field, Fields, Schema, TimeUnit}` (all in `src/arrow/schema.rs`), and every one of those is defined in `arrow-schema` and merely re-exported by `arrow::datatypes`. There's no `arrow-array` / `RecordBatch` / compute-kernel usage anywhere in the crate, so moving to the full `arrow` crate pulls a much heavier dependency into the otherwise-lean spec layer for no functional gain.

Could we keep it as `arrow-schema` (with an `arrow-schema = "59"` workspace dep)? The same applies to `datafusion-iceberg-sql`, which also only touches `arrow::datatypes`. Not blocking — just keeps the spec layer minimal.

**Future note: `SpawnedReader` for `DataFileReader`**

Nice migration off the deprecated `ParquetObjectReader` — implementing `AsyncFileReader` directly is exactly what the deprecation note recommends. That same note also points at parquet's `SpawnedReader` for running the reader's I/O on a dedicated runtime. Not needed for this PR, but worth keeping in mind as a follow-up if we ever want to isolate object-store I/O from CPU-bound parquet decode; it would wrap `DataFileReader` at the construction site in `read.rs` rather than change the reader itself.

gruuya:
> The spec crate only references arrow::datatypes::{DataType, Field, Fields, Schema, TimeUnit} (all in src/arrow/schema.rs), and every one of those is defined in arrow-schema and merely re-exported by arrow::datatypes. There's no arrow-array / RecordBatch / compute-kernel usage anywhere in the crate, so moving to the full arrow crate pulls a much heavier dependency into the otherwise-lean spec layer for no functional gain.

Oh i see the intention now, sorry, let me revert this.

gruuya:
Ok, reverted the arrow-schema removal.

Also added a TODO for `DataFileReader` to consider wrapping it with the `SpawnedReader` down the line.

JanKaul:
Thanks a lot!

PR #400, opened 2026-09-17
Support Arrow Map schemas in Iceberg writes

osipovartem:
## Summary
- convert Arrow `Map` fields to Iceberg `MapType`
- enforce the Iceberg requirement that map keys are non-null
- assign stable field IDs to map keys and values when Arrow metadata is absent
- cover Arrow/Iceberg map round trips and ID assignment

This makes schema conversion symmetric: Iceberg maps could already be represented as Arrow maps, but Arrow maps could not be converted back for writes.

The conversion is schema-level only and does not add per-row work.

## Validation
- `cargo +1.95.0 test -p iceberg-rust-spec arrow::schema::tests` (19 passed)
- `cargo +1.95.0 clippy -p iceberg-rust-spec --all-targets -- -D warnings`
- `rustfmt +1.95.0 --check --edition 2021` on the changed file


JanKaul:
LGTM, thanks!

PR #401, opened 2026-09-17
Support deletion-only overwrite commits

osipovartem:
## Summary

Allow `TableTransaction::overwrite` to commit when it removes existing data files without adding replacement files.

Previously, an overwrite with an empty `data_files` vector returned before processing `files_to_overwrite`. The selected files therefore remained live and no snapshot was committed.

This change adds a deletion-only manifest-list path that:

- copies unaffected manifests unchanged
- returns affected manifests for the existing filtering pipeline
- does not require partition bounds or a selected manifest for new files
- keeps the existing replacement-overwrite path unchanged

The new regression test writes a partitioned table, removes one partition with no replacement files, verifies that a new overwrite snapshot is committed, checks the summary's remaining row count, and reads only the surviving rows.

## Validation

- `cargo +1.95.0 test -p iceberg-rust --test overwrite_test` (3 passed)
- `cargo +1.95.0 clippy -p iceberg-rust --all-targets -- -D warnings`
- `git diff --check`

Developed with AI assistance.


JanKaul:
LGTM, thanks!

PR #402, opened 2026-09-17
Make REST access delegation configurable

osipovartem:
## Summary
- add a configurable `Configuration::access_delegation` value for `X-Iceberg-Access-Delegation`
- preserve the existing `vended-credentials` behavior by default in both `Configuration::default()` and `ConfigurationBuilder`
- use the configured value consistently for both `loadTable` and `createTable`
- allow catalogs to select `remote-signing` or explicitly disable access delegation with `None`
- cover both default construction paths, header generation, and the explicit opt-out

The generated REST API already exposes this header for `createTable` and accepts arbitrary headers for `loadTable`. Keeping the policy in `Configuration` lets clients choose the delegation mode for their catalog without hard-coding it at individual call sites.

## Compatibility and performance

Existing callers continue to request `vended-credentials` by default. The setting only affects REST request header construction and does not affect table scan or write hot paths.

## Verification
- `cargo +1.95.0 test -p iceberg-rest-catalog access_delegation_header_is_enabled_by_default -- --nocapture`
- `cargo +1.95.0 clippy -p iceberg-rest-catalog --all-targets -- -D warnings`
- `cargo +1.95.0 fmt --package iceberg-rest-catalog -- --check`


JanKaul:
Thanks for this! One request before merging: let's make access delegation on by default rather than opt-in.

Right now the field defaults to None, which changes existing behavior — the loadTable fallback currently always sends X-Iceberg-Access-Delegation: vended-credentials, and this PR would silently stop that for anyone who doesn't set the new field. That's a behavior regression for current deployments relying on vended credentials.

Could we default access_delegation to Some("vended-credentials") in both the Default impl and the #[builder(default = ...)]? That keeps the feature (deployments can still choose remote-signing or None) while preserving today's behavior for everyone who doesn't touch the setting. The API change to Configuration is fine.

Two small follow-ups if you make that change:
- The createTable call currently passes None; with an on-by-default value it would start sending the header too. That's the consistent behavior this PR is going for, just worth calling out.
- The access_delegation_header_is_opt_in test should be renamed/updated to assert the default-on behavior.

osipovartem:
Addressed in c64315ac. `access_delegation` now defaults to `Some("vended-credentials")` in both `Configuration::default()` and `ConfigurationBuilder`, preserving the existing load-table behavior. The create-table path receives the same configured default, explicit `None` still disables the header, and the test now covers both default construction paths plus the opt-out. Focused unit tests and `clippy --all-targets -D warnings` pass locally.

JanKaul:
LGTM, thanks!

PR #403, opened 2026-09-17
Return the written row count from the DataFusion sink

osipovartem:
## Summary
- return the number of rows written by `IcebergDataSink::write_all` instead of always returning zero
- derive the result from the already-produced Iceberg `DataFile.record_count` metadata
- reject negative record counts and checked-add overflow
- assert the row count returned by a DataFusion `INSERT`

`DataSink::write_all` returns the number of written rows, which DataFusion exposes as the DML result. Returning zero makes successful Iceberg inserts report an incorrect affected-row count.

## Performance

The implementation performs one checked addition per output data file after those files have already been written. It does not rescan input batches, materialize rows, or add work to the per-row write path.

## Verification
- `cargo +1.95.0 test -p datafusion_iceberg table::tests::test_datafusion_table_insert`
- `cargo +1.95.0 clippy -p datafusion_iceberg --all-targets -- -D warnings`


JanKaul:
LGTM, thanks!

PR #404, opened 2026-09-17
Support REST catalog table rename

osipovartem:
## Summary
- add `Catalog::rename_table` with a default `NotSupported` implementation
- implement atomic table rename through the Iceberg REST Catalog endpoint
- migrate the cached object store from the source identifier to the destination
- extend the existing Apache REST fixture integration test to verify source removal, destination visibility, and the loaded table identifier

The REST client already generates the `tables/rename` endpoint and request model, but the public catalog trait did not expose it. A default implementation keeps existing catalog implementations source-compatible and lets catalogs opt in only when they can provide correct rename semantics.

## Performance

Rename performs one REST request plus O(1) cache-key migration. Table scan and write paths are unchanged.

## Verification
- `cargo +1.95.0 test -p iceberg-rest-catalog test_create_update_drop_table --no-run`
- `cargo +1.95.0 clippy -p iceberg-rest-catalog --all-targets -- -D warnings`

The Docker-backed integration test compiles locally; runtime execution requires the Apache REST fixture via `/var/run/docker.sock` and is left to CI.


JanKaul:
LGTM, thanks!

PR #405, opened 2026-09-17
Support Iceberg decimals with 38-digit precision

osipovartem:
## Summary

- replace the 96-bit `rust_decimal` representation with `fastnum::D128` so `Value::Decimal` can represent the full Iceberg `decimal(38, s)` range
- preserve exact signed i128 mantissas and scales across Iceberg binary values, Parquet statistics, DataFusion pruning, and scalar statistics
- encode decimal values using the minimum-length big-endian two's-complement representation required by the Iceberg specification
- construct decimals directly from fixed-width integer parts, without temporary strings, parsing, or heap allocation

Iceberg allows decimal precision up to 38, while the previous 96-bit mantissa could not represent the maximum positive and negative values.

## API note

This changes the public inner type of `Value::Decimal` from `rust_decimal::Decimal` to the exported `iceberg_rust_spec::spec::decimal::Decimal` alias (`fastnum::D128`). Callers constructing or inspecting decimal values should use the new helper functions in `spec::decimal`.

## Verification

- `cargo +1.95.0 test -p iceberg-rust-spec --lib`
- `cargo +1.95.0 test -p datafusion_iceberg converts_precision_38_decimal_bound_to_datafusion`
- `cargo +1.95.0 test -p datafusion_iceberg --test roundtrip_types`
- `cargo +1.95.0 clippy -p iceberg-rust-spec -p datafusion_iceberg --all-targets -- -D warnings`

Spec reference: https://iceberg.apache.org/spec/#primitive-types


iceberg-rust-spec/src/spec/decimal.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+/// Returns the signed unscaled value.
+#[must_use]
+pub fn decimal_mantissa(decimal: &Decimal) -> i128 {
+    let magnitude = decimal
+        .digits()
+        .to_u128()
+        .expect("an Iceberg decimal has at most 38 digits");
+    let magnitude = i128::try_from(magnitude).expect("38 decimal digits fit in i128");
```
JanKaul: `i128::try_from(...).expect(...)` panics for D128 values above i128::MAX (~1.7e38), which is below D128::MAX (~3.4e38). Nothing validates precision ≤ 38 before this point, and it's a public spec helper — can we return Result<i128, Error> instead of panicking? Same for the expects in decimal_from_i128_with_scale.

JanKaul (commented):
Generally looks good to me. I have one small comment about the `.expects`

iceberg-rust-spec/src/spec/decimal.rs
osipovartem: Fixed in 99cf26f1. Both `decimal_mantissa` and `decimal_from_i128_with_scale` now return `Result` and their callers propagate the error (or conservatively emit null pruning statistics). Decimal-to-ByteBuf serialization now encodes the `U128` coefficient directly as minimal two's-complement bytes, so it remains infallible without narrowing through `i128`. I added regression coverage for scales outside `i32`, coefficients above `i128::MAX`, and byte-for-byte equivalence with the existing `i128` encoding across signed boundary values.

JanKaul:
LGTM, thanks!

PR #406, opened 2026-09-17
Support Iceberg v3 row lineage for appends

osipovartem:
## Summary

- model Iceberg v3 row-lineage metadata on snapshots, manifest-list entries, and data files
- honor format-version=3 when creating table metadata
- assign contiguous first-row-id ranges to every unassigned data manifest when writing V3 manifest lists, advancing by added_rows_count + existing_rows_count
- assign copied pre-upgrade manifests as well as newly written manifests while preserving existing assignments
- materialize inherited per-file row IDs while reading committed manifests before entry filtering
- preserve explicit per-file IDs when reading uncommitted manifests that have not received a manifest-level assignment yet
- keep the V2 manifest-list Avro schema unchanged and add field ID 520 only to the V3 schema
- deserialize pre-field V3 data files and pre-upgrade snapshots through forward-compatible defaults

This PR intentionally supports append paths only. Replace and Overwrite remain explicitly rejected for V3 tables because rewriting those manifests safely requires preserving inherited row IDs at file granularity.

## Correctness and performance

Assignment and inheritance operate once per manifest or data-file metadata entry, never per table row. Existing V3 manifests are copied into the new manifest list instead of being read and rewritten. All manifest-list serialization paths share one assignment and validation helper, and the V3 Avro schema remains cached behind OnceLock. V1/V2 paths retain their existing schema and behavior.

Commit validation rejects negative or overlapping ranges, permits valid forward ranges, checks overflow, and updates next-row-id atomically with the snapshot update. Reader inheritance distinguishes committed pre-upgrade manifests from uncommitted manifests and runs before sequence filtering.

## Verification

- cargo +1.95.0 test -p iceberg-rust-spec --lib --quiet
- cargo +1.95.0 test -p iceberg-rust --lib --quiet
- V1/V2/V3 manifest-list round trips
- worked row-ID assignment example and upgraded existing-only manifests
- mixed reader inheritance, committed/uncommitted null behavior, and overflow tests
- V2-to-V3 append integration test covering manifest and per-file inherited IDs
- cargo +1.95.0 clippy -p iceberg-rust-spec -p iceberg-rust --all-targets -- -D warnings

Spec reference: https://iceberg.apache.org/spec/#row-lineage


JanKaul:
## Review: completing the assignment rules + making the replace/overwrite transition additive

Thanks — the metadata modeling here (field 142/520 placement, `SnapshotV3` serde, commit validation) matches the spec, and `next_row_id = first_row_id + added_rows` in `apply_table_updates` is exactly right: it keeps `next-row-id` above every assigned row ID even when a forward gap is admitted by the `>=` check.

That said, three append-only shortcuts in the writer internals diverge from the spec's general assignment rules in ways that only surface once replace/overwrite/upgrade paths exist — and they'd require reworking this code rather than extending it. Since the fixes are small, I'd like to land them in this PR so the follow-up is purely additive.

### 1. Assignment rule in `append_new_manifest` (manifest_list.rs)

Current:
```rust
if added_rows > 0 {
    manifest.first_row_id = Some(*next_row_id);
    *next_row_id += added_rows;            // added_rows_count only
}
```

The spec's rule ([First Row ID Assignment](https://iceberg.apache.org/spec/#first-row-id-assignment)) is: assign whenever a **data manifest has a null `first_row_id`** — no row-count condition — and advance by **`added_rows_count + existing_rows_count`**:

- The `added_rows > 0` guard skips manifests that must be assigned. An upgraded table's carried-over manifest can have `added = 0, existing > 0`, and the spec is explicit: "Any null (unassigned) `first_row_id` must be assigned via inheritance, even if the data file is existing."
- The added-only advance under-reserves ID space once a manifest can contain unassigned *existing* files (rewrites, upgrades). The spec's recommended increment is `added_rows_count + existing_rows_count`, which is safe in all cases.

**Proposal:** extract the rule into a small `RowIdAssigner` (or free fn) owning the counter:
```rust
/// Spec "First Row ID Assignment": data manifests without a first_row_id are
/// assigned the counter, which advances by added_rows_count + existing_rows_count.
/// Existing assignments and delete manifests are untouched.
fn assign(&mut self, manifest: &mut ManifestListEntry) -> Result<(), Error>
```
Observable behavior for the append-only scope doesn't change (new manifests have `existing = 0`), but the rule is then final and every future path can reuse it.

### 2. Copied-through manifests bypass assignment

In the V3 branch of `from_existing` (and the select helpers), preserved manifests go straight to `writer.append_ser(manifest)`. Per the spec, "when writing the manifest list, **all data manifests** must be assigned a `first_row_id`" — including manifests copied forward from a pre-upgrade table, which still carry null.

**Proposal:** route the copy loop through the same `assign` call, and add an invariant check when serializing a V3 manifest list: any data manifest still lacking a `first_row_id` at that point is an error. That guarantees upgraded-table backfill can never be silently skipped.

### 3. Reader-side inheritance is absent (and it's the enabler for the future rewrite PR)

The spec requires readers to materialize per-file IDs ([First Row ID Inheritance](https://iceberg.apache.org/spec/#first-row-id-inheritance)): within a manifest that has a `first_row_id`, each data file with a null `first_row_id` is assigned the manifest's value plus the running sum of `record_count` of the preceding null-id files. This is also what makes rewrites cheap later: if IDs materialize at manifest-read time, a rewrite path just reads a committed manifest and writes the now-explicit values through — no separate inheritance code.

**Proposal:** add the helper to `iceberg-rust-spec` and apply it in `datafiles()` (`table/mod.rs:306-371`) alongside the existing sequence-number inheritance:
```rust
pub struct FirstRowIdInheritance { ... }
impl FirstRowIdInheritance {
    /// - manifest has first_row_id  -> assign nulls, advance by record_count (skip DELETED)
    /// - committed manifest, null   -> force all entry ids to null (pre-upgrade)
    /// - uncommitted manifest, null -> preserve entry ids untouched
    pub fn apply(&mut self, entry: &mut ManifestEntry)
}
```
Three details worth encoding now:
- Apply **before** any entry filtering (e.g. the sequence-range filter), so skipped entries still advance the running sum.
- **DELETED entries are neither assigned nor advance the counter** — writers must persist explicit IDs when marking entries deleted, so a null-id DELETED entry only occurs pre-upgrade, where its rows have no IDs.
- The **uncommitted-preserve mode**: reading back a not-yet-committed manifest (null `first_row_id` because assignment happens at manifest-list write) must not clobber explicit IDs on EXISTING entries. This is the hook that later lets manifest merging work without touching this code again.

### 4. Avro forward-compat: `#[serde(default)]` on `DataFileV3.first_row_id`

The manifest-list struct got `#[serde(default)]` but the data-file struct didn't. This codebase reads manifests schema-less (`apache_avro::from_value`), so V3 manifests written before this field existed will fail to deserialize. Same pre-existing gap as `referenced_data_file`/`content_offset`/`content_size_in_bytes` (143-145) — worth fixing all four here.

### Suggested tests

- Assignment: the spec's worked example (existing manifest keeps 925; added manifests get 1000/1125/1225 with mixed added/existing counts); upgraded-table manifest (`added = 0, existing > 0, first_row_id = None`) gets assigned; delete manifests stay null.
- Inheritance: mixed manifest (null-id files, explicit-id file, DELETED entry, delete file) against the expected running sum; committed-null manifest → IDs forced null; entries filtered by sequence range still advance the sum.
- Round-trip: V3 manifest written with the pre-field schema reads back (`serde(default)`).

### Explicitly out of scope (follow-up PR)

Lifting the `NotSupported` guard for Replace/Overwrite, restoring manifest merging for V3 (deleting the copy-everything branch), and `_row_id`/`_last_updated_sequence_number` materialization in datafusion. With the two helpers above landed, all of those become additive.

osipovartem:
Addressed in b0658d97. Thanks for the detailed review.

1. Added a shared RowIdAssigner. It assigns every unassigned V3 data manifest and advances by added_rows_count + existing_rows_count, with required/non-negative count and overflow validation. Existing assignments are preserved; delete manifests are forced to null.
2. Copied-through manifests, selected manifests, overwrite selectors, and newly written manifests now all serialize through the same append_manifest assignment/invariant path. A V3 data manifest cannot be serialized without first_row_id.
3. Added FirstRowIdInheritance in iceberg-rust-spec with assigned, committed-pre-upgrade, and uncommitted-preserve modes. The table reader applies it before sequence filtering. Explicit IDs do not advance the counter; DELETED entries and delete files neither inherit nor advance it.
4. Added serde defaults for first_row_id and the three deletion-vector pointer fields on DataFileV3, plus an Avro round-trip test using the pre-field schema.

Tests include the worked assignment example, upgraded manifests with existing-only rows, delete manifests, mixed inheritance, committed/uncommitted null behavior, overflow, schema forward compatibility, and a V2-to-V3 append integration test that verifies manifest and per-file IDs.

Validation:
- cargo +1.95.0 test -p iceberg-rust-spec --lib --quiet: 152 passed
- cargo +1.95.0 test -p iceberg-rust --lib --quiet: 160 passed
- cargo +1.95.0 clippy -p iceberg-rust-spec -p iceberg-rust --all-targets -- -D warnings: passed

JanKaul:
LGTM, thanks

PR #411, opened 2026-09-22
feat: map Iceberg string to Arrow Utf8View

gruuya:
Switch to Utf8View as the default Iceberg string representation.

Without this and the companion changes, the following test panics with `Arrow error: Offset overflow error: 2147483648`
```rust
#[path = "df_common/mod.rs"]
mod df_common;

use df_common::{boot_df_stack, execute_sql, total_rows};

/// 36 000 x 64 KiB = 2.25 GiB of string data. Every row holds the *same*
/// value, so Parquet dictionary-encodes it down to a few KiB on disk -- only
/// the decoded Arrow arrays are large.
const ROWS: usize = 36_000;
const WIDTH: usize = 64 * 1024;

#[tokio::test]
async fn string_column_over_2gib_survives_a_blocking_operator() {
    let ctx = boot_df_stack().await;

    execute_sql(&ctx, "CREATE SCHEMA warehouse.wide").await;
    execute_sql(
        &ctx,
        "CREATE EXTERNAL TABLE warehouse.wide.t (id BIGINT NOT NULL, payload STRING) \
         STORED AS ICEBERG LOCATION '/warehouse/wide/t'",
    )
    .await;
    execute_sql(
        &ctx,
        &format!(
            "INSERT INTO warehouse.wide.t \
             SELECT value, repeat('a', {WIDTH}) FROM generate_series(1, {ROWS})"
        ),
    )
    .await;

    // `count(*) OVER ()` is an unbounded window frame, so WindowAggExec
    // concatenates the entire input -- `payload` included -- into one batch.
    let batches = execute_sql(
        &ctx,
        "SELECT payload, count(*) OVER () FROM warehouse.wide.t",
    )
    .await;

    assert_eq!(total_rows(&batches), ROWS);
}
```

Besides, `Utf8View` has long since been regarded as the best/default choice (over `Utf8`), see e.g. https://github.com/apache/datafusion/pull/13101


JanKaul:
Thanks for the PR. Makes sense to make the shift

PR #413, opened 2026-09-22
Preserve row lineage during v3 overwrites

osipovartem:
## Summary

- preserve Iceberg v3 row lineage during overwrite and manifest rewrites
- validate exact live-file overwrite targets and reject delete manifests/files
- preserve `Deleted` entries plus snapshot and sequence attribution across later appends
- handle schema and partition-spec evolution without mixing historical and current manifest schemas
- normalize partition values and optional data-file bounds using only Iceberg-legal type promotions
- add optimistic reference requirements for first and subsequent commits

This builds on the row-lineage append support merged in #406.

## Correctness

- chained `FLOAT` to `DOUBLE` schema promotion is covered across partial overwrite, append, and a second overwrite
- historical partition specs, snapshots without `schema_id`, and evolved schemas are decoded against the correct writer schema
- overwritten files remain deleted after later appends and cannot be resurrected through manifest reuse
- replacement delete files and incompatible optional metrics fail closed or are discarded as required
- manifest iterator errors are propagated instead of being silently ignored

## API note

`AssertRefSnapshotId.snapshot_id` is now `Option<i64>` so a first-snapshot commit can represent that the target reference must remain absent. This is an intentional source-level API change.

## Performance

- candidate decoders are cached by schema/spec identity
- unchanged values, including same-scale decimals and strings, use no-conversion fast paths
- manifest rewrites use bounded concurrency of eight tasks
- no row-level Arrow or Parquet processing is added to table scans

## Validation

- all 22 transaction operation tests pass
- all 163 active `iceberg-rust-spec` tests pass
- focused manifest-list, schema-evolution, partition-evolution, chained-promotion, and unchanged-bounds regressions pass
- `cargo +1.94.0 clippy -p iceberg-rust -p iceberg-rust-spec --all-targets -- -D warnings`
- `git diff --check`
- independent read-only review covering correctness, regressions, performance, API compatibility, and missing tests: **APPROVED**

The equivalent implementation in the Embucket fork was also validated against a Snowflake-managed Iceberg v3 table. Snowflake read the resulting UPDATE, DELETE, and MERGE commits, and its `CHANGES` table function paired update rows by preserved row ID.


osipovartem:
Independent read-only contribution review completed: no blocking correctness, regression, performance, or test-coverage issues found. APPROVED. Non-blocking API note: the public snapshot_id field changes from i64 to Option<i64>; the compatibility impact should remain called out in release notes.

